PR #2, opened 2018-02-13
Wrap hook call in subshell

Ealdwulf:
Currently hooks are executed in the same shell. This means that if the hook script executes `exit`, subsequent hooks are not run. The man page doesn't document this behavior - indeed, it says all hooks are run. Wrapping in a subshell should avoid this. 

This is a behavior change. If it is anticipated that any users utilise this behavior, then it might not be possible to make this change so simply.

### DIFF
--- a/hooks/dhcpcd-run-hooks.in
+++ b/hooks/dhcpcd-run-hooks.in
@@ -346,6 +346,6 @@ do
 		esac
 	done
 	if [ -f "$hook" ]; then
-		. "$hook"
+		( . "$hook" )
 	fi
 done


rsmarples:
I would rather the current behaviour is explicitly documented so that custom enter/exist hooks can affect the variables passed to other hooks.
I know for sure some users rely on this behaviour.

Ealdwulf:
That's a pity.

rsmarples:
This is now documented in d531f8b063c23d6fa62a6f0c5b5e69e849da4600 with a futher commit to fix a spelling mistake.

PR #8, opened 2019-09-07
check UDP packet length before checksumming it in valid_udp_packet()

maciejsszmigiero:
dhcpcd sometimes crashes on Linux when it reads a truncated UDP packet,
causing valid_udp_packet() to read past the buffer when checksumming it.
This issue can be fixed by adding explicit packet length checks to this
function.

The BPF filter code loaded into the OS kernel checks the received packet
length, but it looks like reading a truncated packet from the kernel is
still possible.

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -3250,7 +3250,7 @@ valid_udp_packet(void *packet, size_t plen, struct in_addr *from,
 		.ip_dst = ip->ip_dst
 	};
 	size_t ip_hlen;
-	uint16_t ip_len, uh_sum;
+	uint16_t ip_len, udp_len, uh_sum;
 	struct udphdr *udp;
 	uint32_t csum;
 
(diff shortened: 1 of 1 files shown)


rsmarples:
Thanks for the excellent patch and commentary. This fixes a recently reported crash I was unable to solve and is a better fix from another submitter.

I've adjusted your patch slightly to move the length check before the don't checksum check as well because we don't want to skip it for sure! I've given you credit in commit c9fc5a5b7

Many thanks!

maciejsszmigiero:
Thank you for your kind words.

I see that you dropped the check for ip_hlen. Is it safe to do so?
My worry is that in the next line the code calculates a checksum of up to ip_hlen bytes
(I know it is limited to 64 bytes but still it seems otherwise unchecked).

rsmarples:
Totally my bad. I have no idea how that got lost!
Added in c8491e1

PR #10, opened 2019-10-19
Fix building on systems with musl

pbui:
musl has its own definition of struct ethhdr, so only include netinet/if_ether.h on systems with GLIBC.  For the ARPHDR constants, we must include linux/if_arp.h instead.

This patch was used downstream in https://github.com/void-linux/void-packages/pull/15616.

### DIFF
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -46,10 +46,18 @@
 
 #include <arpa/inet.h>
 #include <net/if.h>
-#include <netinet/if_ether.h>
 #include <netinet/in_systm.h>
 #include <netinet/in.h>
 #include <net/route.h>
+
+/* musl has its own definition of struct ethhdr, so only include
+ * netinet/if_ether.h on systems with GLIBC.  For the ARPHRD constants,
+ * we must include linux/if_arp.h instead. */
+#if defined(__GLIBC__)
+#include <netinet/if_ether.h>
+#else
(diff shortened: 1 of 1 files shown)


src/if-linux.c
```diff
@@ -46,7 +46,11 @@
 
 #include <arpa/inet.h>
 #include <net/if.h>
+#if defined(__GLIBC__)
```
rsmarples: Could you move thism block to after the network headers and before errno.h with whitespace and a comment to say it's needed for musl?

Thanks

pbui: Sure thing... I've made the request changes.  Let me know if you need any other modifications.

src/if-linux.c
```diff
@@ -46,11 +46,17 @@
 
 #include <arpa/inet.h>
 #include <net/if.h>
-#include <netinet/if_ether.h>
 #include <netinet/in_systm.h>
 #include <netinet/in.h>
 #include <net/route.h>
 
+/* Needed to support compiling with musl */
```
rsmarples: Please change this comment to your initial commit comment has it has more value:

```
musl has its own definition of struct ethhdr, so only include
netinet/if_ether.h on systems with GLIBC.
For the ARPHDR constants, we must include linux/if_arp.h instead.
```

pbui: Ok, I have changed the comment.

PR #12, opened 2020-01-01
Linux: fix RA time unit confusion

andersk:
The RA times are [provided in milliseconds](https://tools.ietf.org/html/rfc4861#section-4.2), but commit 569051c8aa8fc297eb8edb7bd228e0fd353d30c1 (dhcpcd-8.1.3~18) “Linux: prefer ms RA times” incorrectly scaled them by an extra factor of 1000 before writing them to `retrans_time_ms` and `base_reachable_time_ms`. Instead, write the values in milliseconds directly to the `*_ms` files, and correctly convert to jiffies or seconds for the other files if necessary.

Fixes NixOS/nixpkgs#76710.

### DIFF
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -1948,20 +1948,20 @@ if_applyra(const struct ra *rap)
 		error = -1;
 
 	snprintf(path, sizeof(path), "%s/%s/retrans_time_ms", p_neigh, ifname);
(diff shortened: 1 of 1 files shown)


rsmarples:
@andersk during merging this into dhcpcd-8, I think that converting to large timers to jiffies could overflow on 32-bit systems, so I've just removed the support for really old kernels in 2b18b85b2b0c455920b285c7069c6e0c2bfc8d0e

PR #13, opened 2020-04-03
use TEMP_PREFERRED/VALID_LIFETIME for OpenBSD

sthen:
Shortly after dhcpcd started using these constants, they were dropped from OpenBSD kernel headers as they're only used by rad(8). Switch to dhcpcd's own constants instead, fixing build on recent -current.


fobser:
It's actually slaacd(8) not rad(8) :)

PR #14, opened 2020-05-05
fix buffer size in getgrnam_r

mpu:


src/if-options.c
```diff
@@ -2095,7 +2095,7 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 			logerr(__func__);
 			return -1;
 		}
-		while ((i = getgrnam_r(arg, &grpbuf, p, (size_t)l, &grp)) ==
+		while ((i = getgrnam_r(arg, &grpbuf, p, (size_t)dl, &grp)) ==
```
rsmarples, line 2098: The cast to size_t can be removed with this fix.

mpu: Done.

PR #15, opened 2020-05-05
Do not report an error if control_group is already set

mpu:
After an update I noticed that dhcpcd reported an error "controlgroup: wheel: not found". Looking into the code it seems that this is due to the new privsep code not moving /etc/group in the chroot. This patch is a quick fix for the error that simply checks if control_group was already parsed successfully.

Here is what the error looked like on my machine:
```
[root@light ~]# dhcpcd wlp3s0
dhcpcd-9.0.2 starting
controlgroup: wheel: not found
wlp3s0: connected to Access Point `Maison'
...
```

### DIFF
--- a/src/if-options.c
+++ b/src/if-options.c
@@ -2120,7 +2120,8 @@ invalid_token:
 			return -1;
 		}
 		if (grp == NULL) {
-			logerrx("controlgroup: %s: not found", arg);
+			if (!ctx->control_group)
+				logerrx("controlgroup: %s: not found", arg);
 			free(p);
 			return -1;
 		}
@@ -2129,7 +2130,8 @@ invalid_token:
 #else
 		grp = getgrnam(arg);
 		if (grp == NULL) {
-			logerrx("controlgroup: %s: not found", arg);
+			if (!ctx->control_group)
+				logerrx("controlgroup: %s: not found", arg);
 			return -1;
 		}
 		ctx->control_group = grp->gr_gid;


src/if-options.c
```diff
@@ -2120,7 +2120,9 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 			return -1;
 		}
 		if (grp == NULL) {
-			logerrx("controlgroup: %s: not found", arg);
+			if (!ctx->control_group) {
```
rsmarples, line 2123: Can you test against `!= NULL` please? It's more explicit and reads better in my eyes.
Also, the bracnes are not needed as the logerrx is just one line

src/if-options.c
```diff
@@ -2129,7 +2131,9 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 #else
 		grp = getgrnam(arg);
 		if (grp == NULL) {
-			logerrx("controlgroup: %s: not found", arg);
+			if (!ctx->control_group) {
```
rsmarples, line 2134: Same as above.

src/if-options.c
mpu: It actually is of type `gid_t`, so I don't think it'd make sense to compare it to `NULL`. In [control.c](https://github.com/rsmarples/dhcpcd/blob/904b0f7288119a7ddcd9aa862530746031cb9e36/src/control.c#L248) you also test it without giving an explicit constant to compare.

src/if-options.c
mpu: I now dropped the braces.

PR #18, opened 2020-09-09
Script execution debug log: Replace back-tick by single quote

bbczeuz:
This PR replaces the initial back-tick before the script name by a single quote. Simplifies parsing of the output log

Before:
```
dhcpcd-9.1.4 starting
no such user dhcpcd
DUID 00:04:72:01:1e:01:52:c2:11:cb:9d:49:d3:65:99:10:ec:5c
dummy0: executing `/usr/libexec/dhcpcd-run-hooks' PREINIT
dummy0: executing `/usr/libexec/dhcpcd-run-hooks' CARRIER
dummy0: IAID 26:55:d3:72
dummy0: delaying IPv4 for 0.7 seconds
dummy0: reading lease `/var/db/dhcpcd/dummy0.lease'
dummy0: soliciting a DHCP lease
dummy0: sending DISCOVER (xid 0xc74ba8ba), next in 3.1 seconds
[...]
```

After:
```
dhcpcd-9.1.4 starting
no such user dhcpcd
DUID 00:04:72:01:1e:01:52:c2:11:cb:9d:49:d3:65:99:10:ec:5c
dummy0: executing '/usr/libexec/dhcpcd-run-hooks' PREINIT
dummy0: executing '/usr/libexec/dhcpcd-run-hooks' CARRIER
dummy0: IAID 26:55:d3:72
dummy0: delaying IPv4 for 0.7 seconds
dummy0: reading lease '/var/db/dhcpcd/dummy0.lease'
dummy0: soliciting a DHCP lease
dummy0: sending DISCOVER (xid 0xc74ba8ba), next in 3.1 seconds
[...]
```

(Compare lines 4, 5 & 8)

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -1164,7 +1164,7 @@ read_lease(struct interface *ifp, struct bootp **bootp)
 		logdebugx("reading standard input");
 		sbytes = read(fileno(stdin), buf.buf, sizeof(buf.buf));
 	} else {
-		logdebugx("%s: reading lease `%s'",
+		logdebugx("%s: reading lease '%s'",
 		    ifp->name, state->leasefile);
 		sbytes = dhcp_readfile(ifp->ctx, state->leasefile,
 		    buf.buf, sizeof(buf.buf));
@@ -2314,7 +2314,7 @@ dhcp_bind(struct interface *ifp)
 	state->state = DHS_BOUND;
 	if (!state->lease.frominfo &&
 	    !(ifo->options & (DHCPCD_INFORM | DHCPCD_STATIC))) {
-		logdebugx("%s: writing lease `%s'",
+		logdebugx("%s: writing lease '%s'",
 		    ifp->name, state->leasefile);
 		if (dhcp_writefile(ifp->ctx, state->leasefile, 0640,
 		    state->new, state->new_len) == -1)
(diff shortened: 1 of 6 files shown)


bbczeuz:
Don't merge yet, there are more similar issues. Is there some intention behind this?

bbczeuz:
Actually, there are still 28 such cases. What's the intention behind this?

bbczeuz:
Commit 127c06e fixes the remaining cases, just in case you want to merge this change.

rsmarples:
@bbczeuz I think the intention was that gcc at the time - or something else I was using many years ago, did the same.
I think I would be happier if we just removed them and maybe prefix with : .... like so

```
dummy0: executing '/usr/libexec/dhcpcd-run-hooks' CARRIER
dummy0: reading lease `/var/db/dhcpcd/dummy0.lease'
```

becomes

```
dummy0: executing: /usr/libexec/dhcpcd-run-hooks CARRIER
dummy0: reading lease: /var/db/dhcpcd/dummy0.lease
```

Thoughts?

bbczeuz:
I don't really mind. Quotes suggest a string, but we don't do any escaping anyhow, so it doesn't matter. Do you want me to update the PR to the colon-version (and rebase?)?

rsmarples:
It's all a string :)
A lot of BSD tools behave like this, so I fine with it.
If people want to be nutty enough to put a space in the script filename for example, that's up to them.
Note that the SSID used in lease files *is* escaped so would look like

```
/var/db/dhcpcd/dummy0\ a\ spaced\ ssid.lease
```

Although it might escape using octal rather than a space. I've not checked that in a while.

Sure, go head and do it and we'll see about merging it.

rsmarples:
I've decided to do this myself as I need to get a release out soon.
Thanks anyway for the patch.

PR #19, opened 2020-11-03
fix if_getnetworknamespace matching bug.

artisdom:
skip netns if either "st_dev" or "st_ino" not matching.


rsmarples:
Ohhh, nice find, thanks!

PR #33, opened 2021-04-23
[WIP] Linux: set broadcast flag by default for ipvlan interfaces.

### LINKED ISSUE #32: Automatically enable requesting broadcast reply on `ipvlan`
The `dhcpcd.conf` manual states:
```
     broadcast
             Instructs the DHCP server to broadcast replies back to the client.  Normally this is only set for non-Ethernet interfaces, such as FireWire and InfiniBand.  In most cases, dhcpcd will set this automatically.

```

Would it be possible to do this for `ipvlan` interfaces as well?

ido:
The patch uses ethtool's `SIOCETHTOOL` `ioctl` to obtain the driver name of the interface, and turns on the broadcast flag by default when it detects an `ipvlan` interface.  This does not yet differentiate between L2, L3, or L3S mode.

@rsmarples we can use this `ethtool` method to detect most other device types, like `bridge` and `tuntap`, without reading `/sys`.  I might refactor those after this is cleaned up.

TODO:
- [ ] Modify DHCP IAID so that `ipvlan` interfaces do not generate an IAID conflict: `ipvlan0: IAID conflicts with one assigned to eth0`
- [x] Clean up patch before merging.

Fixes #32

### DIFF
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -42,6 +42,7 @@
 #include <linux/if_vlan.h>
 #include <linux/filter.h>
 #include <linux/netlink.h>
+#include <linux/ethtool.h> /* for ipvlan detection */
 #include <linux/sockios.h>
 #include <linux/rtnetlink.h>
 
@@ -309,11 +310,54 @@ if_init(struct interface *ifp)
 	return if_writepathuint(ifp->ctx, path, 1) == -1 ? -1 : 0;
 }
 
+/* Returns number of bytes written to driver, else 0 (error or indeterminate). */
+static size_t
+if_get_driver(struct interface *ifp, char *driver, const size_t driverlen)
+{
+	struct ethtool_drvinfo drvinfo = { .cmd = ETHTOOL_GDRVINFO };
+	struct ifreq ifr = { .ifr_data = (void *)&drvinfo };
+
+	strlcpy(ifr.ifr_name, ifp->name, sizeof(ifr.ifr_name));
+	if (ioctl(ifp->ctx->pf_inet_fd, SIOCETHTOOL, &ifr) != 0) {
+		logerr("%s: SIOCETHTOOL ifname=%s", __func__, ifp->name);
+		return 0; /* 0 means error or indeterminate driver name */
+	}
+	return strlcpy(driver, drvinfo.driver, driverlen);
+}
+
+static bool
+if_cmp_driver(struct interface *ifp, const char *driver)
+{
+	char ifdriver[FIELD_SIZEOF(struct ethtool_drvinfo, driver)];
+	size_t n = if_get_driver(ifp, ifdriver, sizeof(ifdriver));
+
+	if (n == 0) {
+		logerr("%s: if_get_driver ifname=%s", __func__, ifp->name);
+		return false;
+	}
+	if (strncmp(ifdriver, driver, n) == 0)
+		return true;
+	return false;
+}
+
+static bool
+if_ipvlan(struct interface *ifp)
+{
+	if (if_cmp_driver(ifp, "ipvlan"))
+		return true;
+	return false;
+}
+
 int
 if_conf(struct interface *ifp)
 {
 	char path[sizeof(SYS_LAYER2) + IF_NAMESIZE];
 	int n;
+
+	/* Set broadcast flag for ipvlan interfaces.
+	 * XXX: move this out to dhcpcd if needed on other platforms. */
+	if (if_ipvlan(ifp))
+		ifp->options->options |= DHCPCD_BROADCAST;
 
 	/* Some qeth setups require the use of the broadcast flag. */
 	snprintf(path, sizeof(path), SYS_LAYER2, ifp->name);
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -338,10 +338,9 @@ if_cmp_driver(struct interface *ifp, const char *driver)
 		logerr("%s: if_get_driver ifname=%s", __func__, ifp->name);
 	} else if (n == 0) {
 		logerr("%s: driver name empty ifname=%s", __func__, ifp->name);
-	} else if (n > 0) {
-		if (strncmp(ifdriver, driver, n) == 0)
-			return true;
 	}
+	if ((n >= 0) && (strncmp(ifdriver, driver, (size_t)n) == 0))
+		return true;
 	return false;
 }
 
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -310,8 +310,11 @@ if_init(struct interface *ifp)
 	return if_writepathuint(ifp->ctx, path, 1) == -1 ? -1 : 0;
 }
 
-/* Returns number of bytes written to driver, else 0 (error or indeterminate). */
-static size_t
+/* Return values:
+ *   -1 = ioctl error
+ *    0 = no driver name or driver name indeterminate
+ *   >0 = length of driver name written to provided buffer. */
+static ssize_t
 if_get_driver(struct interface *ifp, char *driver, const size_t driverlen)
 {
 	struct ethtool_drvinfo drvinfo = { .cmd = ETHTOOL_GDRVINFO };
@@ -320,23 +323,25 @@ if_get_driver(struct interface *ifp, char *driver, const size_t driverlen)
 	strlcpy(ifr.ifr_name, ifp->name, sizeof(ifr.ifr_name));
 	if (ioctl(ifp->ctx->pf_inet_fd, SIOCETHTOOL, &ifr) != 0) {
 		logerr("%s: SIOCETHTOOL ifname=%s", __func__, ifp->name);
-		return 0; /* 0 means error or indeterminate driver name */
+		return -1; /* -1 means ioctl error */
 	}
-	return strlcpy(driver, drvinfo.driver, driverlen);
+	return (ssize_t)strlcpy(driver, drvinfo.driver, driverlen);
 }
 
 static bool
 if_cmp_driver(struct interface *ifp, const char *driver)
 {
 	char ifdriver[FIELD_SIZEOF(struct ethtool_drvinfo, driver)];
-	size_t n = if_get_driver(ifp, ifdriver, sizeof(ifdriver));
+	ssize_t n = if_get_driver(ifp, ifdriver, sizeof(ifdriver));
 
-	if (n == 0) {
+	if (n == -1) {
 		logerr("%s: if_get_driver ifname=%s", __func__, ifp->name);
-		return false;
+	} else if (n == 0) {
+		logerr("%s: driver name empty ifname=%s", __func__, ifp->name);
+	} else if (n > 0) {
+		if (strncmp(ifdriver, driver, n) == 0)
+			return true;
 	}
-	if (strncmp(ifdriver, driver, n) == 0)
-		return true;
 	return false;
 }
 
--- a/src/common.h
+++ b/src/common.h
@@ -63,6 +63,10 @@
 #define ROUNDUP4(a)		(1 + (((a) - 1) |  3))
 #define ROUNDUP8(a)		(1 + (((a) - 1) |  7))
 
+#ifndef FIELD_SIZEOF
+#define FIELD_SIZEOF(t, f) sizeof(((t *)0)->f)
+#endif
+
 /* Some systems don't define timespec macros */
 #ifndef timespecclear
 #define timespecclear(tsp)      (tsp)->tv_sec = (time_t)((tsp)->tv_nsec = 0L)
--- a/src/dhcpcd.8.in
+++ b/src/dhcpcd.8.in
@@ -651,7 +651,7 @@ of a randomly generated number.
 .It Fl J , Fl Fl broadcast
 Instructs the DHCP server to broadcast replies back to the client.
 Normally this is only set for non-Ethernet interfaces,
-such as FireWire and InfiniBand.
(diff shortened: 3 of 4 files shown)


src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
 }
 
+/* @maxlen should be greater than or equal to 32; see ethtool_drvinfo. */
+static int
+if_get_driver(const char *ifname, char *driver, const size_t maxlen)
+{
+	struct ifreq ifr;
+	struct ethtool_drvinfo drvinfo;
+	int fd;
+	if (!ifname || !*ifname)
```
rsmarples, line 320: dhcpcd strives to follow NetBSD KNF style.

As such a newline after variable decs is required.
We prefer explicit NULL checks, but it's not mandatory.
Also, NULL is not the NULL character.
Lastly, if returning an error, set errno.

```
if (ifname == NULL || *ifname == '\0') {
    errno = EINVAL;
```

Infact checking if the first character in ifname is the NULL character is redunant as the ioctl will return a suitable
error for us.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
 	return if_writepathuint(ifp->ctx, path, 1) == -1 ? -1 : 0;
 }
 
+/* @maxlen should be greater than or equal to 32; see ethtool_drvinfo. */
+static int
+if_get_driver(const char *ifname, char *driver, const size_t maxlen)
+{
+	struct ifreq ifr;
```
rsmarples, line 317: You can avoid the memset below by setting a member to 0.
`struct ifreq ifr = { .ifr_flags = 0 };`

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+	fd = socket(AF_INET, SOCK_STREAM, 0);
+	if (fd == -1) {
+		logerr("if_get_driver: failed socket creation: %s", strerror(errno));
+		return -1;
+	}
+
+	memset(&ifr, 0, sizeof(ifr));
+	strlcpy(ifr.ifr_name, ifname, sizeof(ifr.ifr_name));
+	drvinfo.cmd = ETHTOOL_GDRVINFO;
+	ifr.ifr_data = (void *)&drvinfo;
```
rsmarples, line 332: ifr_data is void anyway, so the cast shouldn't be required.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+	if (ioctl(fd, SIOCETHTOOL, &ifr) < 0) {
```
rsmarples, line 333: ioctl on Linux always returns -1 on error.
Some ioctls can return negative values as success, so an explicit -1 test is generally better.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+		const int err = errno;
+		logerr("if_get_driver: failed ethtool ioctl on interface %s: %s",
```
rsmarples, line 335: logerr() will append strerror(errno) and preserve errno.
logerrx() does not.

`logerr("%s: SIOCETHTOOL ifname=%s", __func__, ifname);`

might be better.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+			ifname, strerror(errno));
+		close(fd);
+		return -1;
+	}
+	close(fd);
+	return strlcpy(driver, drvinfo.driver,
+		MIN(sizeof(drvinfo.driver), maxlen));
```
rsmarples, line 342: MIN here is redundant because the size of the drvinfo.driver property is irrelevant to the size of the driver variable.

`strlcpy(driver, drvinfo.driver, maxlen);`

is fine.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
 	return if_writepathuint(ifp->ctx, path, 1) == -1 ? -1 : 0;
 }
 
+/* @maxlen should be greater than or equal to 32; see ethtool_drvinfo. */
+static int
+if_get_driver(const char *ifname, char *driver, const size_t maxlen)
```
rsmarples, line 315: maxlen is a poor name as more parameters could be added. driverlen would be better as it's more descriptive as to it's purpose.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+	}
+	close(fd);
+	return strlcpy(driver, drvinfo.driver,
+		MIN(sizeof(drvinfo.driver), maxlen));
+}
+
+static bool
+if_ipvlan(const char *ifname)
+{
+	char driver[32]; /* sizeof(ethtool_drvinfo.driver) = 32 */
```
rsmarples, line 348: Should use `IF_NAMESIZE` rather than 32 as `IF_NAMESIZE` is defined by the system headers.
A newline after the variable decs would be nice.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+	if (if_get_driver(ifname, driver, sizeof(driver)) < 0) {
+		logerr("if_ipvlan: failed to get driver name %s.", ifname);
```
rsmarples, line 350: You can use __func__ with logerr like so
`logerr("%s: failed to get driver name %s", __func__, ifname);`
Generally debug lines don't end with a full stop.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+		return false;
+	}
+	return (strncmp(driver, "ipvlan", sizeof(driver)) == 0);
```
rsmarples, line 353: the extra braces around strncmp are superfluous.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+}
+
 int
 if_conf(struct interface *ifp)
 {
 	char path[sizeof(SYS_LAYER2) + IF_NAMESIZE];
 	int n;
 
+	/* Set broadcast flag for ipvlan interfaces. */
+	if (if_ipvlan(ifp->name)) {
```
rsmarples, line 363: Braces for a single statement are superfluous.
No biggie though, that's personal preference.

src/if-linux.c
```diff
@@ -309,12 +310,60 @@ if_init(struct interface *ifp)
...
+		ifp->options->options |= DHCPCD_BROADCAST;
```
rsmarples, line 364: The only downside of setting this here rather than in dhcpcd.c is that if any other OS gains ipvlan functionality then this logic might be forgotten about.

That's why I created stub functions in if-bsd.c and if-sun.c that just return false.

rsmarples (changes requested):
Mainly comments on style.
Otherwise looks good!

deliciouslytyped:
Please remember to change the documentation as well!

src/if-linux.c
ido: Doing as you suggest, but wouldn't advise it in general because padding would not be zeroed, which when applied to data that passes over a socket, can leak information. (More relevant in the privileged to unprivileged direction.)

src/if-linux.c
ido: I did it this way to minimize the amount of times SIOCETHTOOL is called in case we end up using this method to detect many other driver types: bridge, tuntap, macvlan, etc. are all detectable this way instead of using the sysfs method currently in use.  (I think SIOCETHTOOL might even be preferred to sysfs for driver detection.)

src/if-linux.c
ido: Assumes drvinfo.driver is properly null-terminated, but fair point.

src/if-linux.c
ido: It's `__caddr_t` (`char *`) on my system, which generates a compiler warning without this cast.

src/if-linux.c
ido: Added an `XXX` to remind myself of this, but AFAIK `ipvlan` is unique to Linux.  (There are other OSes with L3IPVLAN interfaces, but they behave differently.)

src/if-linux.c
```diff
@@ -309,12 +310,70 @@ if_init(struct interface *ifp)
...
+	struct ethtool_drvinfo drvinfo;
+	int fd;
+	size_t n = 0;
+
+	fd = socket(AF_INET, SOCK_STREAM, 0);
+	if (fd == -1) {
+		logerr("%s: socket ifname=%s", __func__, ifname);
+		goto eexit;
+	}
+	memset(&ifr, 0, sizeof(ifr));
```
rsmarples, line 328: This can now be removed.
BTW, you can clarify the structure is correctly zeroed by using Valgrind or compiling with the address or memory sanitizer provided by your compiler.
Although you might need to disable privsep, entirely or just the sandbox + chroot + privdrop to get it to work.

src/if-linux.c
```diff
@@ -309,12 +310,70 @@ if_init(struct interface *ifp)
 	return if_writepathuint(ifp->ctx, path, 1) == -1 ? -1 : 0;
 }
 
+/* Returns number of bytes written to driver, 0 on error.
+ * @driverlen note that sizeof(ethtool_drvinfo.driver) = 32 */
+static size_t
+if_get_driver(const char *ifname, char *driver, const size_t driverlen)
+{
+	struct ifreq ifr = { .ifr_flags = 0 };
+	struct ethtool_drvinfo drvinfo;
```
rsmarples, line 319: This structure needs initialisation or it may pass unintended data to the ioctl.

src/if-linux.c
```diff
@@ -309,12 +310,70 @@ if_init(struct interface *ifp)
...
+eexit:
+	if (fd != -1)
+		close(fd);
+	return n;
+}
+
+static bool
+if_cmp_driver(const char *ifname, const char *driver)
+{
+	char ifdriver[32]; /* see ethtool_drvinfo.driver declaration */
```
rsmarples, line 346: Does this work? (untested)

`char ifdriver[sizeof(((struct ethtool_drvinfo *)0)->driver)]`

src/if-linux.c
```diff
@@ -309,12 +310,70 @@ if_init(struct interface *ifp)
...
+ * @driverlen note that sizeof(ethtool_drvinfo.driver) = 32 */
+static size_t
+if_get_driver(const char *ifname, char *driver, const size_t driverlen)
+{
+	struct ifreq ifr = { .ifr_flags = 0 };
+	struct ethtool_drvinfo drvinfo;
+	int fd;
+	size_t n = 0;
+
+	fd = socket(AF_INET, SOCK_STREAM, 0);
```
rsmarples, line 323: This may not work in privsep.
Please use if_ioctl, see here for an example:
https://github.com/NetworkConfiguration/dhcpcd/blob/master/src/if-linux.c#L505

rsmarples: Actually that's not enough.
I think we need to move `if_indirect_ioctl` from `if-bsd.c`  to `if.c`
https://github.com/NetworkConfiguration/dhcpcd/blob/master/src/if.c#L149

Also move `PS_IOCTLINDIRECT` and friends from `privsep-bsd.c` to`privsep-root.c`
That's quite an invasive change though.

src/if-linux.c
rsmarples: A quick glance at the structure shows it should have no padding on all arches so it could be initialised in the same way.

src/if-linux.c
ido: Yes, that's FIELD_SIZEOF in the Linux kernel source, I'll add it to `common.h` as part of this patch. 👍

src/if-linux.c
ido: IIUC, SIOCETHTOOL with ETHTOOL_GDRVINFO should not require root on most systems, and that's the only kernel we need to support this on (for now).  (i.e. `ethtool -i $INTERFACE` should work for any user.)

Currently, `if_vlanid(...)` uses `ifp->ctx->pf_inet_fd` directly.  I could do that instead of creating my own socket, would that be enough for now if we're only supporting `ipvlan` on Linux for this patch?

ido:
I squashed the commits into one patch.  So far, this sets broadcast.  I might have a solution for setting unique IAIDs soon.

src/if-linux.c
rsmarples: > Currently, `if_vlanid(...)` uses `ifp->ctx->pf_inet_fd` directly. I could do that instead of creating my own socket, would that be enough for now if we're only supporting `ipvlan` on Linux for this patch? That simplifies things quite a bit.

Yes, that's fine.

src/if-linux.c
```diff
@@ -309,12 +310,55 @@ if_init(struct interface *ifp)
...
+static size_t
+if_get_driver(struct interface *ifp, char *driver, const size_t driverlen)
+{
+	struct ethtool_drvinfo drvinfo = { .cmd = ETHTOOL_GDRVINFO };
+	struct ifreq ifr = { .ifr_data = (void *)&drvinfo };
+
+	strlcpy(ifr.ifr_name, ifp->name, sizeof(ifr.ifr_name));
+	if (ioctl(ifp->ctx->pf_inet_fd, SIOCETHTOOL, &ifr) != 0) {
+		logerr("%s: SIOCETHTOOL ifname=%s", __func__, ifp->name);
+		return 0; /* 0 means error or indeterminate driver name */
```
rsmarples, line 323: I would change the function return type to ssize_t and return -1 here to distinguish between an ioctl error and zero length driver name.

PR #38, opened 2021-05-04
configure: test -a|o is not POSIX

concatime:
Same as https://github.com/linux-pam/linux-pam/pull/311 and https://github.com/karelzak/util-linux/pull/1216.

I had `test: too many arguments` when building dhcpcd using [sbase](https://core.suckless.org/sbase/).

Please check that I broke nothing.

Thank you.

### DIFF
--- a/configure
+++ b/configure
@@ -194,7 +194,7 @@ done
 _which()
 {
 	x="$(which "$1" 2>/dev/null)"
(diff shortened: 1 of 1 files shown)


rsmarples:
LGTM, needs testing on OpenBSD as that builds statically by default.

PR #42, opened 2021-07-30
Fix parsing of hex/octal escapes in strings

cmcqueen:
This affects the parsing of strings containing hex or octal escapes.

Prior to this change, if parameter `sbuf` was `NULL`, it would miscalculate the length of the string `l`. Eg `"test\x31"` or `"test\0061"` would be calculated as length 4.

### DIFF
--- a/src/if-options.c
+++ b/src/if-options.c
@@ -321,9 +321,10 @@ parse_str(char *sbuf, size_t slen, const char *str, int flags)
 						break;
 					c[i] = *str++;
 				}
-				if (c[1] != '\0' && sbuf) {
+				if (c[1] != '\0') {
 					c[2] = '\0';
-					*sbuf++ = (char)strtol(c, NULL, 16);
+					if (sbuf)
+						*sbuf++ = (char)strtol(c, NULL, 16);
 				} else
 					l--;
 				break;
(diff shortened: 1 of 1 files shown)


cmcqueen:
I discovered this issue when trying to use the configuration `vendclass`, specifying some binary data. I found the binary data was being truncated.

rsmarples:
Could you post an example where it was failing please?

cmcqueen:
I gave three examples above. What more would you like?

rsmarples:
Sorry, didn't see those.

src/if-options.c
```diff
@@ -335,11 +336,12 @@ parse_str(char *sbuf, size_t slen, const char *str, int flags)
...
 					c[i] = *str++;
 				}
-				if (c[2] != '\0' && sbuf) {
+				if (c[2] != '\0') {
 					i = (int)strtol(c, NULL, 8);
 					if (i > 255)
 						i = 255;
-					*sbuf ++= (char)i;
+					if (sbuf)
+						*sbuf ++= (char)i;
```
rsmarples, line 344: Could you change this from `*sbuf ++=` to `*sbuf++ =` to match the rest of the code please? Yes, I know the mistake is originally mine.

cmcqueen: Sure, can do. I've force-pushed that change. Is that okay?

rsmarples: It's fine, thanks.

PR #43, opened 2021-07-30
Enforce proper alignment of serialized 'struct cmsghdr' inside 'psm->data'

mcingel:
The 'struct cmsghdr', referenced by 'msg_control', requires 'long' alignment. Right now the 'msg_control' is placed in data buffer directly after 'msg_name', that in some cases contains IPv6 header (**28_B**). In this case it violates expected alignment on 64-bit architectures according to 'CMSG_ALIGN' from socket.h


src/privsep.c
```diff
@@ -828,9 +828,11 @@ ps_unrollmsg(struct msghdr *msg, struct ps_msghdr *psm,
     const void *data, size_t len)
 {
 	uint8_t *datap, *namep, *controlp;
+	unsigned long cmsg_padlen = psm->ps_controllen ? \
```
rsmarples, line 831: Should be socklen_t maybe?

src/privsep.c
```diff
@@ -885,16 +889,24 @@ ps_sendpsmmsg(struct dhcpcd_ctx *ctx, int fd,
...
 		psm->ps_controllen = (socklen_t)msg->msg_controllen;
 
 		iovp->iov_base = msg->msg_name;
 		iovp->iov_len = msg->msg_namelen;
 		iovp++;
+
+		cmsg_padlen = msg->msg_controllen ? \
+			CMSG_ALIGN(msg->msg_namelen) - msg->msg_namelen : 0;
+		iovp->iov_len = cmsg_padlen;
+		iovp->iov_base = cmsg_padlen != 0 ? padding : NULL;
```
rsmarples, line 904: Do we need to assert that cmsg_padlen <= sizeof(padding) ?

src/privsep.c
```diff
@@ -1015,19 +1029,20 @@ ps_sendcmdmsg(int fd, uint16_t cmd, const struct msghdr *msg)
 	}
 
 	if (msg->msg_controllen != 0) {
-		if (msg->msg_controllen > dl)
+		if (msg->msg_controllen + cmsg_padlen > dl)
 			goto nobufs;
+		p += cmsg_padlen; /* create padding */
```
rsmarples, line 1034: Should the memory for padding be zeroed?

src/privsep.c
mcingel: done

src/privsep.c
mcingel: done

src/privsep.c
mcingel: Added, but not really necessary. The `cmsg_padlen` can be only in range `<0; sizeof(long) - 1>`. It is safe as long as `padding` occupies  memory of `long` type.

src/privsep.c
```diff
@@ -1004,6 +1017,8 @@ ps_sendcmdmsg(int fd, uint16_t cmd, const struct msghdr *msg)
 		{ .iov_base = data, .iov_len = 0 },
 	};
 	size_t dl = sizeof(data);
+	socklen_t cmsg_padlen = msg->msg_controllen ? \
+		(socklen_t)(CMSG_ALIGN(msg->msg_namelen) - msg->msg_namelen) : 0;
```
rsmarples, line 1021: This is going over 80 char width which I try to enforce. It's also duplicated above and somewhat duplicated again on line 831.
Maybe convert it to a macro around line 84?

src/privsep.c
```diff
@@ -1027,7 +1047,7 @@ ps_sendcmdmsg(int fd, uint16_t cmd, const struct msghdr *msg)
 	if (psm.ps_datalen > dl)
 		goto nobufs;
 
-	iov[1].iov_len = psm.ps_namelen + psm.ps_controllen + psm.ps_datalen;
+	iov[1].iov_len = psm.ps_namelen + psm.ps_controllen + psm.ps_datalen + cmsg_padlen;
```
rsmarples, line 1050: Can you move `+ cmsg_padlen` to a newline with 4 space ident please to avoid going over 80 chars?

rsmarples (changes requested):
This now looks good. Just need to address some formatting issues.

src/privsep.c
mcingel: done

src/privsep.c
mcingel: ok, done

src/privsep.c
```diff
@@ -828,9 +832,11 @@ ps_unrollmsg(struct msghdr *msg, struct ps_msghdr *psm,
     const void *data, size_t len)
 {
 	uint8_t *datap, *namep, *controlp;
+	socklen_t cmsg_padlen = CALC_CMSG_PADLEN(psm->ps_controllen,\
```
rsmarples, line 835: Why the trailing \ character?

mcingel: fixed

rsmarples:
I did some more formatting fixes and added CMSG_ALIGN for non Linux platforms.

Thanks for the patch!

Issue #47, opened 2021-08-17
Infinite loop with DHCP Inform option


klanmiko:
Hi. This issue relates to symptoms found in [this RPi thread](https://www.raspberrypi.org/forums/viewtopic.php?t=265516)

In my own testing, on dhcpcd 9.4.0, running dhcpcd from the command line like:`dhcpcd -B -s 10.10.10.3/24 eth0` does cause the syslog to be filled with messages like: 
Aug 16 14:56:49 meraki dhcpcd[361]: eth0: received approval for 10.10.10.3
Aug 16 14:56:49 meraki dhcpcd[361]: eth0: received approval for 10.10.10.3
Aug 16 14:56:49 meraki dhcpcd[361]: eth0: received approval for 10.10.10.3

In addition, the device sends repeated DHCPINFORM (and ARP announcements) at a high frequency.

I did some debugging and it seems that with the first DHCP ACK, `dhcp_bind` calls `ipv4_applyaddr` which adds the address to the kernel. There is a handler for the kernel response `dhcp_handleifa` which calls `dhcp_inform` again, continuing the cycle.

Another phenomena I have noticed is that future DHCP ACKS are processed both by the newly opened UDP socket and the bpf filter, which keeps getting closed by dhcp_bind and reopened by dhcp_inform

I will try to contribute a fix but I would appreciate help solving this in a way that won't introduce a regression, since I am not familiar with dhcpcd's code.

seamusdemora:
Hello - I've just found this while searching the cause of the 8/second log entries I see when using the recommended `inform <ipv4 addr>` option (instead of `static ip`; i.e.

```
# edit /etc/dhcpcd.conf; added 2 lines
# interface wlan0
# inform 192.168.1.199/24
```

After save & reboot:

```
$ journalctl -f | grep wlan0
Mar 26 07:15:17 raspberrypi0w dhcpcd[568]: wlan0: received approval for 192.168.1.199
Mar 26 07:15:17 raspberrypi0w dhcpcd[568]: wlan0: received approval for 192.168.1.199
Mar 26 07:15:17 raspberrypi0w dhcpcd[568]: wlan0: received approval for 192.168.1.199
Mar 26 07:15:17 raspberrypi0w dhcpcd[568]: wlan0: received approval for 192.168.1.199
Mar 26 07:15:17 raspberrypi0w dhcpcd[568]: wlan0: received approval for 192.168.1.199
[...]

PR #48, opened 2021-08-17
dhcpcd-9.4.0: Fix infinite DHCP INFORM messages

klanmiko:
When an address is configured on the interface, do not send a DHCP
INFORM message if the interface is already bound (ack already received)

The second issue is that dhcp_inform resets the lease information, and
may drop the existing server address.

#47

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -2619,6 +2619,7 @@ dhcp_inform(struct interface *ifp)
 	    &ia->addr, &ia->mask);
 	if (state->offer_len) {
 		dhcp_new_xid(ifp);
+		get_lease(ifp, &state->lease, state->offer, state->offer_len);
 		send_inform(ifp);
 	}
 }


src/dhcp.c
```diff
@@ -2619,7 +2619,6 @@ dhcp_inform(struct interface *ifp)
 	    &ia->addr, &ia->mask);
 	if (state->offer_len) {
 		dhcp_new_xid(ifp);
-		get_lease(ifp, &state->lease, state->offer, state->offer_len);
```
rsmarples, line 2622: Why was this removed? Doesn't look right.

klanmiko: This line resets the server address, forces a broadcast out. Should we broadcast or use the open UDP socket? I need to do additional testing of what side effects deleting this line might have.

geki-yaba:
Thanks for this patch. This helps with a mesh setup of APs. I often have "reconnects", which puts me offline for some time - quite annoying. Dunno what it internally does?! Maybe AP transition.. And with this patch, I no longer have these down times.

Just a heads up for usefulness of this patch. :)

PR #51, opened 2021-08-23
Update man pages for default metric changes

peterfab9845:
In commit 3547ed19, the default metric offsets were changed. This PR updates the man pages accordingly.


rsmarples:
I would lead with the 2000 for wireless as that applies to all OS.
I would then go on to say that if the OS supports route metrics than a futher offset of 1000000 will be applied for IPv4LL routes and 2000000 when the interface is roaming.

peterfab9845:
Ah, I had missed that the offsets are cumulative. Fixed.

PR #73, opened 2021-11-13
Add epair(4) to the ignored interface types list

woodsb02:
The FreeBSD epair(4) interface type can be thought of as a virtual patch cable.
The epair interfaces come in pairs, with network traffic being passed between the 2 pairs.
The basic intent is to provide connectivity between two virtual network stack instances.
This interface type does not typically require an IP address, as it is passing traffic
between other virtual networks/interfaces which obtain their own IP address.
Therefore, add this interface type to the list that will be ignored by dhcpcd.

### DIFF
--- a/src/if-bsd.c
+++ b/src/if-bsd.c
@@ -104,6 +104,7 @@
  * just won't work without explicit configuration. */
 static const char * const ifnames_ignore[] = {
 	"bridge",
+	"epair",	/* Virtual patch cable */
 	"fwe",		/* Firewire */
 	"fwip",		/* Firewire */
 	"tap",


Crest:
DHCP on `epair(4)` interfaces can be very useful to provide autoconfiguration to vnet enabled jails. It's a policy question. Please don't break the mechanism required to implement different policies.

Still I can see the argument to disable them by default. Will it be enough to add `epair*` to `allowinterfaces` to reenable them? Would it be enough to provide an explanation of this imo reasonable policy and `denyinterfaces epair*` through the FreeBSD ports default configuration?

A more complex and FreeBSD specific default (that should still be overridable) would be to only ignore epair interfaces if the jail id is zero (aka dhcpcd isn't jailed) as epair interfaces are probably the most common way to connect vnet enabled jails to the outside world (e.g. by adding the other end on to a software bridge on the jail host also containing a physical NIC) and user who decides to run dhcpcd inside a vnet enabled they probably expects DHCP to work over `epair*` interfaces.

woodsb02:
Valid point if you want DHCP inside a jail. For some reason that never occurred to me, since I only use static IP jails (doh). My issue was the interfaces on the host were also sending DHCP queries.

Yes, they still work if you add them to the dhcpcd.conf file using the “allowinterfaces” configuration setting.

I like the idea to enable them by default if jail id is non-zero. Do you think you could come up with a patch for this?

Otherwise, probably the best idea is as you said - revert this change and instead update the FreeBSD port to include a denyinterfaces statement in the default dhcpcd.conf file with a comment.

What’s not clear to me is whether the allowinterfaces / denyinterfaces config settings work in the same way as this ifnames_ignore function, where the pattern can match the interface group not just the interface name. This is important because some jail management tools rename the epair interfaces (e.g. bastillebsd renames them from epair0a/b to e0a/b_jailname). If the group isn’t matched the default config would have to be something like “denyinterfaces e*[ab]_*”

Crest:
It should be enough to check the `security.jail.jailed` sysctl (0 = not jailed, 1 = jailed, all other values = assume jailed as well). Since a process either starts as jailed or has to jail itself and the attack surface of reading a integer sysctl and comparing it to zero is negligible I recommend reading the sysctl before entering capability mode and caching the result. If the child processes are re`exec()`ed to reduce attack surface and can't access the cached sysctl value directly it can be exported as environment variable those mere presence is easy to detected by following `environ` and doesn't require parsing. I'm not familiar with the dhcpcd codebase so cleaning up any patch I could offer is probably more work than it's worth. If you're willing to review a patch anyways I'm willing to take a shot at it.

PR #85, opened 2022-02-09
Fix handling of hostname argument

semihalf-ostruszka-andrzej:
"hostname" argument (-h) is a simple string and should never be
interpreted as a hex string.  It is also kept as a null terminated
string, so when we parse something that is meant to be null terminated
string do not attempt to parse it as a HW addressa.  The same applies to
"script" (-c) argument.

Signed-off-by: Andrzej Ostruszka <amo@semihalf.com>

### DIFF
--- a/src/if-options.c
+++ b/src/if-options.c
@@ -264,7 +264,7 @@ parse_str(char *sbuf, size_t slen, const char *str, int flags)
 			str++;
 			end = p;
 		}
-	} else {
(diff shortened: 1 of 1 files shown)


semihalf-ostruszka-andrzej:
Hello

Has this change been superseded by something else?  I just wanted to check current version and have noticed that this is not present and would like to double check if this is intentional.

With regards
Andrzej

PR #86, opened 2022-02-10
Use a local variable instead of the optind

gotthardp:
The `optind` is a global variable that gets overwritten by `reload_config()`, so the `reconf_reboot()` used a wrong argument count. As a result, renewal of a single interface always caused renewal of all interfaces.

Also, the `optind - 1` at the end of the function feel wrong as the `reconf_reboot()` just processes the list of interface names.

### DIFF
--- a/src/dhcpcd.c
+++ b/src/dhcpcd.c
@@ -1493,7 +1493,7 @@ dhcpcd_handleargs(struct dhcpcd_ctx *ctx, struct fd_list *fd,
 {
 	struct interface *ifp;
 	unsigned long long opts;
-	int opt, oi, do_reboot, do_renew, af = AF_UNSPEC;
+	int opt, oi, oifind, do_reboot, do_renew, af = AF_UNSPEC;
 	size_t len, l, nifaces;
 	char *tmp, *p;
 
@@ -1509,7 +1509,7 @@ dhcpcd_handleargs(struct dhcpcd_ctx *ctx, struct fd_list *fd,
 		return control_queue(fd, UNCONST(fd->ctx->cffile),
 		    strlen(fd->ctx->cffile) + 1);
 	} else if (strcmp(*argv, "--getinterfaces") == 0) {
-		optind = argc = 0;
+		oifind = argc = 0;
 		goto dumplease;
 	} else if (strcmp(*argv, "--listen") == 0) {
 		fd->flags |= FD_LISTEN;
@@ -1572,6 +1572,9 @@ dhcpcd_handleargs(struct dhcpcd_ctx *ctx, struct fd_list *fd,
 		}
 	}
 
+	/* store the index; the optind will change when a getopt get called */
+	oifind = optind;
+
 	if (opts & DHCPCD_DUMPLEASE) {
 		ctx->options |= DHCPCD_DUMPLEASE;
 dumplease:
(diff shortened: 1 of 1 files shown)


src/dhcpcd.c
```diff
@@ -1651,7 +1654,7 @@ dhcpcd_handleargs(struct dhcpcd_ctx *ctx, struct fd_list *fd,
 
 	reload_config(ctx);
 	/* XXX: Respect initial commandline options? */
-	reconf_reboot(ctx, do_reboot, argc, argv, optind - 1);
+	reconf_reboot(ctx, do_reboot, argc, argv, oifind);
```
rsmarples, line 1657: Why remove the -1?

rsmarples, line 1657: Why remove the -1?

gotthardp: Are you really sure the -1 should be there? The `optind` is the index of the first interface name. It is sometimes even set to `0`. The `reconf_reboot` processes interface names. It feels to me that it shouldn't be decreased.

rsmarples: The relevant commit is here: https://github.com/NetworkConfiguration/dhcpcd/commit/8f3a9e56d0a4a1cf7bca9845a9c1b8349e7d6c0d

This imples there was an off by one issue when processing command options via the control socket.

agowa338:
Any update with this?

I just rediscovered this issue. Is this PR still active?

rsmarples:
I haven't had the time to look more into it. See commit 8f3a9e56d0a4a1cf7bca9845a9c1b8349e7d6c0d for why it's currently like it is. If you can provide more rationale, always welcome :)

agowa338:
I'll look into it, but I cannot promise to be able to help with that particular issue as I currently don't understand the logic there.

ipaton1:
The proposed fix is good and commit 8f3a9e5 is incorrect.  The commit message gives no context to the original issue, so it's hard to say for sure but I think it could be due to the way it was tested.

Step back to what @gotthardp says in the PR. `optind` is a _global_ defined in unistd.h therefore it can only be valid while you're using the getopt family of functions on a single argv.  Consider what would happen if you start parsing a second argv halfway through parsing a first, `optind` will get corrupted at the very least.

However `dhcpcd_handleargs` uses it while parsing the command that's been passed through the control socket and towards the end of the function just before the call to `reload_config` optind will point to the interface name that was passed via the control socket. 

Assume a command of the form `dhcpcd -n eth0` and you'll have optind==2  and argv[optind] == 'eth0'

Next `reload_config` is called which shortly calls `add_options(ctx, NULL, ifo, ctx->argc, ctx->argv);` then very quickly into `add_options` there's `optind = 0;` at this point you're re-parsing the commandline from the original invocation of the master dhcpcd process and when you're done with that `optind` will be at the end of the original commandline, potentially in a completely different place from where it would have been when you finished parsing the command line from the control socket. Remember that optind is a global, so it won't get reset when you exit the function.

After we return from `reload_config` we then call `reconf_reboot(ctx, do_reboot, argc, argv, optind - 1);` where this argc/argv is the commandline that was passed through the control socket, but optind is from parsing the argc/argv of the original master process command line.

So lets consider the following. The original master process command line is something like `dhcpcd -f config eth0`
Then the command to send through the control socket is `dhcpcd -n eth0` 
The position of eth0 and hence the eventual optind is different making it look like there's an off-by-one problem and hence the -1 is required.  Something like this seems a likely explanation for why it's there given the abscence of further details that were not added to the commit message.

My normal dhcpcd command line has something like seven arguments resulting in optind-1 = 6 which is well off the end of the argv array from the control socket that just has `-n eth0`

I didn't dig too deeply into `reconf_reboot` but it looks like it won't actually try to dereference argv with an index off the end of the array as optind is likely always going to be greater than argc, but that probably needs verified.

As @gotthardp says this results in dhcpcd running on _all_ interfaces (including lo !) 

There's possibly also further bugs lurking here too, for instance why do we end up trying to do dhcp on the lo interface? 

Anyway, given the explanation above, any chance we can have this merged?

rsmarples:
OK, I cannot really refute this.
I'll merge as is and see if I can recall why I had the -1. It might have been related to this very issue really.

PR #93, opened 2022-03-28
add support for loongarch64

yetist:
LoongArch is a new instruction set of Loongson 3A5000 CPU, you can read the [documents](https://github.com/loongson/LoongArch-Documentation) to get more infomation.

This PR add support for the LoongArch architecture, please review, thanks.


xen0n:
We seemingly doesn't need to wait for seccomp/libseccomp#356, because we're not depending on libseccomp for our seccomp integration -- seccomp is fully supported on the Linux/LoongArch port already.

Anyway this looks good to me -- some people (including me) have already reviewed this at loongarch64/dhcpcd#1 before this submission.

rsmarples:
I don't have the arch, but if people say it works that's fine with me.

PR #121, opened 2022-08-16
On FreeBSD, don't send SIOCSPFXFLUSH_IN6

mekanix:
Unless this portion of code is removed, routes which belong to bridge0 interface (unrelated to the interface being configured) are removed. For example

```
# ifconfig bridge0 inet6 -ifdisabled auto_linklocal fd10:6c79:8ae5:8b92::1
# netstat -rn | grep fd10
fd10:6c79:8ae5:8b92::/64          link#6                        U       bridge0
fd10:6c79:8ae5:8b92::1            link#6                        UHS         lo0
# dhcpcd -B re0 # interupt after a while
# netstat -rn | grep fd10
fd10:6c79:8ae5:8b92::1            link#6                        UHS         lo0
```

I don't know what exactly is wrong, but this at least narrows it down. I don't expect this PR will be accepted in the current state, but I thought PR is the best way to show my findings.

### DIFF
--- a/src/if-bsd.c
+++ b/src/if-bsd.c
@@ -1906,12 +1906,6 @@ if_setup_inet6(const struct interface *ifp)
 		    &ifr, sizeof(ifr)) == -1 &&
 		    errno != ENOTSUP && errno != ENOTTY)
 			logwarn("SIOCSRTRFLUSH_IN6 %d", errno);
-#ifdef SIOCSPFXFLUSH_IN6
-		if (if_ioctl6(ifp->ctx, SIOCSPFXFLUSH_IN6,
-		    &ifr, sizeof(ifr)) == -1 &&
-		    errno != ENOTSUP && errno != ENOTTY)
-			logwarn("SIOCSPFXFLUSH_IN6");
-#endif
 	}
 #endif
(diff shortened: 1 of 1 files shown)


mekanix:
Just to be clear, re0 in this case is not member of bridge0, so they are totally unrelated.

rsmarples:
This needs to be fixed FreeBSD side and will not be accepted in dhcpcd.

From a quick glance, this chunk needs deleting:
https://github.com/freebsd/freebsd-src/blob/e7d02be19d40063783d6b8f1ff2bc4c7170fd434/sys/netinet6/in6.c#L599
down to `aifaddr_out:`

This needs removing too:
https://github.com/freebsd/freebsd-src/blob/41335c6b7f636f9ca59d0afba5728cf90020d6b1/sys/netinet6/in6_ifattach.c#L493

Both parts should be replaced to create a generic prefix route separate from the router handling clode.

mekanix:
As this PR only serves to expose the problem and the problem is being handled on the FreeBSD side, I think this PR fulfilled it's purpose, hence I'm closing it. Thank you for swift response and patches!

PR #145, opened 2022-11-01
arp: ignore invalid ARP probes according to RFC5227

laszloth:
Some network devices can send probes where the source address in the frame does not match the sender in the ARP request, due to firmware bugs or crafted packets.
Without verification this can cause to fail the defense, so we lose the address, then request a new one. This might lead to the DHCP address pool being exhausted and all sorts of problems like frequent IP changes.

A real life example is that some Cisco devices can send ARP probes in bursts, where the frame's source address is clearly invalid, so hosts running dhcpcd release addresses frequently, e.g., Raspberry Pis. Example output:
"invalid ARP probe, sender hw address mismatch (00:be:75:xx:xx:xx, 00:00:00:yy:yy:yy)"

RFC5227, Section 1.1 describes that the two addresses must match in case of ARP probes, so add the extra check and log a warning if we receive an invalid packet like that.

Signed-off-by: Laszlo Toth <laszlth@gmail.com>

### DIFF
--- a/src/arp.c
+++ b/src/arp.c
@@ -232,6 +232,9 @@ arp_packet(struct interface *ifp, uint8_t *data, size_t len,
 	const struct iarp_state *state;
 	struct arp_state *astate, *astaten;
 	uint8_t *hw_s, *hw_t;
+#ifndef KERNEL_RFC5227
+	bool is_probe;
+#endif /* KERNEL_RFC5227 */
 
 	/* Copy the frame header source and destination out */
 	memset(&arm, 0, sizeof(arm));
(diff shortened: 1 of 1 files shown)


src/arp.c
```diff
@@ -232,6 +232,7 @@ arp_packet(struct interface *ifp, uint8_t *data, size_t len,
 	const struct iarp_state *state;
 	struct arp_state *astate, *astaten;
 	uint8_t *hw_s, *hw_t;
+	bool is_probe;
```
rsmarples, line 235: This will need to be guarded by `#ifndef KERNEL_RFC5227` as well I think.

rsmarples:
This took me a while to process and initallly my knee jerk reaction was it's wrong. But I think I now agree with your interpretation.
If you can address the one review comment I'll get this merged into the next release.

laszloth:
> This took me a while to process and initallly my knee jerk reaction was it's wrong. But I think I now agree with your interpretation. If you can address the one review comment I'll get this merged into the next release.

Yeah, it looks like a corner case, but saw a few very similar forum error reports during the initial debugging + some Cisco reports, and the RFC says "MUST", so when I lost my IP again due to this on the rpi I decided to open it upstream.

Issue #146, opened 2022-11-01
dhcpcd incorrectly removes DNS records


ChenQi1989:
Version:
3c458fc7fa4146029a1e4f9e98cd7e7adf03081a (tag: dhcpcd-9.4.1)

Environment:
qemu + slirp network + dhcpcd + systemd-resolved

Behavior:
dhcpcd first correctly configures DNS via resolvconf, then it removes the records, resulting in incorrect /etc/resolv.conf.

Output from 'journalctl | grep 10.0.2'
  Nov 01 05:32:12 qemux86-64 dhcpcd[233]: eth0: rebinding lease of 10.0.2.15
  Nov 01 05:32:12 qemux86-64 dhcpcd[233]: eth0: leased 10.0.2.15 for 86400 seconds
  Nov 01 05:32:12 qemux86-64 dhcpcd[233]: eth0: adding route to 10.0.2.0/24
  Nov 01 05:32:12 qemux86-64 dhcpcd[233]: eth0: adding default route via 10.0.2.2
  **Nov 01 05:32:12 qemux86-64 systemd-resolved[188]: eth0: Bus client set DNS server list to: 10.0.2.3**

I added a few debug messages in dhcpcd-hooks/20-resolv.conf.
Output from 'journalctl | grep qi'
  Nov 01 05:32:11 qemux86-64 dhcpcd-systemd-wrapper[255]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
  Nov 01 05:32:11 qemux86-64 dhcpcd-systemd-wrapper[255]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
  Nov 01 05:32:11 qemux86-64 dhcpcd-systemd-wrapper[264]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
  Nov 01 05:32:11 qemux86-64 dhcpcd-systemd-wrapper[264]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
  Nov 01 05:32:12 qemux86-64 dhcpcd-systemd-wrapper[334]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
  Nov 01 05:32:12 qemux86-64 dhcpcd-systemd-wrapper[334]: [qi]: dhcpcd resolv.conf hook is run, have_resolvconf=true
[...]

PR #150, opened 2022-11-08
hooks/20-resolv.conf: early return instead of remove_resolv_conf

ChenQi1989:
In practice, this script is executed again and again. Except for the a few early ones, the remaining executions all met the condition that there's no configuration. Yet, the DNS server is still there. So instead of removing resolv.conf, we'd better just early return here.

Fixes issue#146.

### DIFF
--- a/hooks/20-resolv.conf
+++ b/hooks/20-resolv.conf
@@ -136,12 +136,11 @@ add_resolv_conf()
 		fi
 	fi
 
-	# If we don't have any configuration, remove it
+	# If we don't have any configuration, early return
 	if [ -z "$new_domain_name_servers" ] &&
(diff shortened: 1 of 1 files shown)


rsmarples:
When a message has no DNS the config needs to be removed for that interface.protocol.
DHCP has no means of saying "Delete these DNS entries". Infact DHCP just says "here are the options" and the long standing assumption is that any prior options not given should be removed.

ChenQi1989:
@rsmarples Thanks Roy. I saw you replied in these two PRs as well as in the https://github.com/NetworkConfiguration/dhcpcd/issues/146. Let's discuss this issue in https://github.com/NetworkConfiguration/dhcpcd/issues/146 to avoid inconvenience.

PR #151, opened 2022-11-09
20-resolv.conf: in case of IPv6 RA, allow for lacking DNS info

ChenQi1989:
https://www.rfc-editor.org/rfc/rfc8106:
"""
These issues are not pressing in dual-stack networks as long as a DNS server is available on the IPv4 side, but they become more critical with the deployment of IPv6-only networks. As a result, this document defines a mechanism based on DNS RA options to allow IPv6 hosts to perform automatic DNS configuration.
"""

According to the above statements, automatic DNS configuration is optional in IPv6 RA, as long as IPv4 has got DNS information. So we should allow for lacking DNS info in case of IPv6 RA.


rsmarples:
Again, like DHCP, RA has no means of removing DNS records *easily*. You can set them to expire early but not many RA routers actually support this.

PR #152, opened 2022-11-11
20-resolv.conf: improve the sitation of working with systemd

ChenQi1989:
systemd's resolvconf implementation ignores the protocol part. See https://github.com/systemd/systemd/issues/25032.

When using 'dhcp server + dns server + dhcpcd + systemd', we get an integration issue, that is dhcpcd runs 'resolvconf -d eth0.ra', yet systemd's resolvconf treats it as eth0. This will delete the DNS information set by 'resolvconf -a eth0.dhcp'.

Fortunately, 20-resolv.conf has the ability to build the resolv.conf file contents itself. We can just pass the generated contents to systemd's resolvconf. This way, the DNS information is not incorrectly deleted. Also, it does not cause behavior regression for dhcpcd in other cases.

### DIFF
--- a/hooks/20-resolv.conf
+++ b/hooks/20-resolv.conf
@@ -11,8 +11,13 @@ nocarrier_roaming_dir="$state_dir/roaming"
 NL="
 "
 : ${resolvconf:=resolvconf}
+
+resolvconf_from_systemd=false
 if command -v "$resolvconf" >/dev/null 2>&1; then
 	have_resolvconf=true
+	if [ $(basename $(readlink -f $(which $resolvconf))) = resolvectl ]; then
+		resolvconf_from_systemd=true
+	fi
 else
 	have_resolvconf=false
 fi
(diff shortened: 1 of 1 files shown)


ChenQi1989:
@rsmarples kindly ping

rsmarples:
The bottom line is that I will not merge this or anything like it into dhcpcd.
Any issues with the system supplied resolvconf should be fixed in resolvconf.

I say this because it's really hard for the script to know which versions of resolvctl work and which don't.
It used to work fine, now it doesn't. I'm sure it will get fixed upstream eventually.

The more important matter is that you are bypassing resolvconf entirely for this corner case and other parts of the system might rely on resolvconf returning correct data rather than `/etc/resolv.conf` being "correct".

ChenQi1989:
@rsmarples 
1. Looking at the git history and its current status, systemd's resolvconf (resolvectl) has never supported the .protocol suffix.
2. I'm not bypassing resolvconf, because the dhcpcd generated contents are supplied to resolvconf. In this way, other parts of the system using resolvconf can continue to work. If I still miss something, please let me know.

Anyway, I understand your opinion and respect your choice.

rsmarples:
OK, so all resolvconf version ever don't actually care about the .protocol suffix. You could do `echo "nameserver 1.2.3.4" | resolvconf -a this_interface.is_usless` and it will work just fine. `interface[.protocol]` is just suggestion. It's just a label, nothing more, nothing less.

If systemd wants to come along and start enforcing otherwise that's a systemd problem, not a dhcpcd problem.

Issue #160, opened 2023-01-03
local subnet classless static routes are treated as host routes


xvuko:
[RFC3443](https://www.rfc-editor.org/rfc/rfc3442.html) allows sending local subnet routes by using 0.0.0.0 gateway.  These routes are treated as host routes and netmask is set to /32.

Relevant part of `decode_rfc3442_rt` function:
```c
/* An on-link host route is normally set by having the
 * gateway match the destination or assigned address */
if (gateway.s_addr == dest.s_addr ||
    (gateway.s_addr == bootp->yiaddr ||
    gateway.s_addr == bootp->ciaddr))
{
	gateway.s_addr = INADDR_ANY;
	netmask.s_addr = INADDR_BROADCAST;
}
```

Both `dest.s_addr` and `bootp->ciaddr` can be set to zero in valid non host routes.

PR #161, opened 2023-01-03
dhcp: Fix classless link local static routes

xvuko:
Fix for #160 

Detecting host routes based on address comparison is not needed in classless routes as /32 mask can be explicitly used. This detecting mechanism did cause issues when gateway was set to 0.0.0.0 (link local routes).


rsmarples:
This might break some old ISP setups, but I cannot fault your logic either.

PR #175, opened 2023-01-24
Add debug logging for system state changes

agowa:
Add logging for reading and writing to system files including /proc


src/ipv6nd.c
```diff
@@ -71,14 +71,49 @@
...
+	uint8_t		nd_opt_route_information_prefix_len;
+	uint8_t 	nd_opt_route_information_reserved_1 : 3; // 3-bit unused fields
+	int8_t		nd_opt_route_information_prf : 2; // 2 bit signed integer
+	uint8_t		nd_opt_route_information_reserved_2 : 3; // 3-bit unused fields
+	uint32_t	nd_opt_route_information_lifetime;
+	/* followed by 128 bit for the prefixes, but only prefix_len bits are valid */
+};
+__CTASSERT(sizeof(struct nd_opt_route_information) == 7);
+#endif
+
```
rsmarples, line 108: Lets move the changes in this file to a new PR please as this is independant of the debugging your adding everywhere.

agowa338: Will do. Most of it was a preparation for (trying to) add RFC 4191 support.
I only included it because of the change to the logic of the switch case in line 1505 (of the original file).

And the last of these three headers isn't currently used anywhere anyway. That one was in error.

igalic:
this looks like a level of detail where you might as well use DTrace for

(or eBPF on Linux these days?)

agowa:
The reason was that dhcpcd was changing system configuration that I had explicitely defined otherwise. So I wasted excessive time trying to understand why sysctl and systemd are buggy. I ran everything with debug output, but dhcpcd didn't log anything, so I was assuming that it didn't change anything and wasted a lot of time trying to debug systemd...

rsmarples:
What sysctls are we talking about here that would cause a service distruption? dhcpcd will change the following IPv6 related tunables:
* disables the kernel from making it's own IPv6 LL addresses because it will make it's own stable private address.
* disables the kernel from doing autoconf or accept_ra because dhcpcd will do it instead
* adjust interface tunables for hop limit, retrans and reachable based on RA received

You can stop dhcpcd from doing this by disabling IPv6 or IPv6RS support in dhcpcd.

Otherwise I'm not sure what else dhcpcd would stamp on and cause a problem?

agowa:
It has been a while, but if I recall correctly, it was an interaction between multiple settings (at least accept_ra and forwarding) that caused the disruption.

The system was used to run containers (I.E. forwarding was enabled). But dhcpcd set accept_ra to 1 (or 0, don't know anymore), whereas I had set it to 2 within a startup task explicitly. Other systems on the same lan were advertising their internal container networks, so that traffic to them was routed directly instead of to the gateway (which filtered it).

PR #178, opened 2023-01-26
Fix incorrect CTASSERT in ND_OPT_DNSSL

agowa338:
The __CTASSERT within ND_OPT_DNSSL is incorrectly referencing nd_opt_rdnss instead of nd_opt_dnss.

The error was undetected until now because both are equal in size...


rsmarples:
Good spot, thanks for the patch!
Luckily it's the same size so it doens't have a functional change.

perkelix:
What is the practical impact of this bug?

agowa338:
@perkelix There is no impact as both are equal in size. So the computer does not care.

We basically have two structs with equal size but different meaning. And within the assert we check one twice instead of each one once.

PR #181, opened 2023-02-01
privsep-linux: fix SECCOMP_AUDIT_ARCH missing ppc64le

chenxy1988:
Behavior: When dhcpcd runs on the ppc64le platform, it would be killed by SIGSYS.

After enabling seccomp debug in privsep-linux, the SIGSYS signal handler shows us the arch is ppc64 not ppc64le, so add ppc64le support in privsep-linux.c

Thanks.


chenxy1988:
Tested and verified on yocto qemuppc64 build.

PR #219, opened 2023-07-01
Fixing Segfault in privsep-root 10.0.1

archcloudlabs:
OS: Arch Linux
Version: dhcpcd 10.0.1

Currently, with dhcpcd 10.0.1 when executing the binary as a non-root user, a segfault occurs when dereferencing values from a struct that is not populated.
error

This PR addresses this issue by checking whether or not the the uid is 0.
fix

### DIFF
--- a/src/privsep-root.c
+++ b/src/privsep-root.c
@@ -962,15 +962,19 @@ ps_root_stop(struct dhcpcd_ctx *ctx)
 	 * log dhcpcd exits because the latter requires the former.
 	 * So we just log the intent to exit.
 	 * Even sending this will be a race to exit. */
-	if (psp) {
-		logdebugx("%s%s%s will exit from PID %d",
-		    psp->psp_ifname,
-		    psp->psp_ifname[0] != '\0' ? ": " : "",
-		    psp->psp_name, psp->psp_pid);
+
+	if (getuid() == 0 ) {
+		if (psp) {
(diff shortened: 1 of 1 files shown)


src/privsep-root.c
```diff
@@ -962,15 +962,19 @@ ps_root_stop(struct dhcpcd_ctx *ctx)
 	 * log dhcpcd exits because the latter requires the former.
 	 * So we just log the intent to exit.
 	 * Even sending this will be a race to exit. */
-	if (psp) {
-		logdebugx("%s%s%s will exit from PID %d",
-		    psp->psp_ifname,
-		    psp->psp_ifname[0] != '\0' ? ": " : "",
-		    psp->psp_name, psp->psp_pid);
+
+	if (getuid() == 0 ) {
```
igalic, line 966: maybe preserve the indentation level by combining these two conditions with an `&&`.

Also extend the comment above explaining why the `getuid()` check is necessary

igalic (changes requested):

rsmarples:
Thanks for the report, the fix you provided can be done a lot better.
Your diagnosis was spot on, but we only need to unlink if the struct is populated in the first place so just testing if the first element is not null is sufficient.

archcloudlabs:
Thanks for the feedback!

PR #223, opened 2023-07-10
Support libcrypto for hmac and sha256

tobhe:
In a recent security review for the Ubuntu package of dhcpcd we found that the packaged version uses the built-in crypto primitives from `compat/crypt`.

We try to keep the number of low-level crypto implementations at a minimum to make it easier for the Ubuntu security team to review them, track eventual bugs as and ensure compliance in certified environments.

This PR adds alternative `compat/crypt` wrappers based on libcrypto and some tests to make sure they work correctly. I made sure that they also work  across different OpenSSL versions and with LibreSSL.

### DIFF
--- a/configure
+++ b/configure
@@ -37,6 +37,7 @@ POLL=
 SMALL=
 SANITIZE=no
 STATUSARG=
+OPENSSL=
 
 DHCPCD_DEFS=dhcpcd-definitions.conf
 
@@ -110,6 +111,8 @@ for x do
 	--with-udev) DEV=yes; UDEV=yes;;
 	--without-udev) UDEV=no;;
 	--with-poll) POLL="$var";;
+	--with-openssl) OPENSSL=yes;;
+	--without-openssl) OPENSSL=no;;
 	--sanitise|--sanitize) SANITIZEADDRESS="yes";;
 	--serviceexists) SERVICEEXISTS=$var;;
 	--servicecmd) SERVICECMD=$var;;
@@ -1350,14 +1353,6 @@ if [ "$FLS64" = yes ]; then
 	echo "#define	HAVE_SYS_BITOPS_H" >>$CONFIG_H
 fi
 
-# Workaround for DragonFlyBSD import
-if [ "$OS" = dragonfly ]; then
-	echo "#ifdef	USE_PRIVATECRYPTO" >>$CONFIG_H
-	echo "#define	HAVE_MD5_H" >>$CONFIG_H
-	echo "#define	SHA2_H			<openssl/sha.h>" >>$CONFIG_H
-	echo "#else" >>$CONFIG_H
-fi
-
 if [ -z "$MD5" ]; then
 	MD5_LIB=
 	printf "Testing for MD5Init ... "
@@ -1387,14 +1382,6 @@ EOF
 	fi
 	echo "$MD5"
 	rm -f _md5.c _md5
-fi
-if [ "$MD5" = no ]; then
-	echo "#include			\"compat/crypt/md5.h\"" >>$CONFIG_H
-	echo "MD5_SRC=	compat/crypt/md5.c" >>$CONFIG_MK
-else
-	echo "MD5_SRC=" >>$CONFIG_MK
-	echo "#define	HAVE_MD5_H" >>$CONFIG_H
-	[ -n "$MD5_LIB" ] && echo "LDADD+=		$MD5_LIB" >>$CONFIG_MK
 fi
 
 if [ -z "$SHA2_H" ]; then
@@ -1481,23 +1468,6 @@ EOF
 		rm -f _sha256.c _sha256
 	fi
 fi
-if [ "$SHA2" = no ]; then
-	echo "#include			\"compat/crypt/sha256.h\"" >>$CONFIG_H
-	echo "SHA256_SRC=	compat/crypt/sha256.c" >>$CONFIG_MK
-else
-	echo "SHA256_SRC=" >>$CONFIG_MK
-	echo "#define	SHA2_H			<$SHA2_H>" >>$CONFIG_H
-	if [ "$SHA2_RENAMED" = yes ]; then
-		echo "#define	SHA256_CTX	SHA2_CTX" >>$CONFIG_H
-		echo "#define	SHA256_Init	SHA256Init" >>$CONFIG_H
-		echo "#define	SHA256_Update	SHA256Update" >>$CONFIG_H
-		echo "#define	SHA256_Final	SHA256Final" >>$CONFIG_H
-	fi
-	[ -n "$SHA2_LIB" ] && echo "LDADD+=		$SHA2_LIB" >>$CONFIG_MK
-fi
-
-# Workarond for DragonFlyBSD import
-[ "$OS" = dragonfly ] && echo "#endif" >>$CONFIG_H
 
 if [ -z "$HMAC" ]; then
 	HMAC_LIB=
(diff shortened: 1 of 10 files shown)


configure
```diff
@@ -1349,6 +1352,51 @@ if [ "$OS" = dragonfly ]; then
...
+int main(void) {
+	return OPENSSL_VERSION_NUMBER;
+}
+EOF
+	if $XCC _openssl.c -lcrypto -o _openssl 2>&3; then
+		OPENSSL=yes
+		MD5=yes
+		HMAC=no
+		echo "#define	WITH_OPENSSL" >>$CONFIG_H
+		echo "LDADD+=		-lcrypto" >>$CONFIG_MK
```
rsmarples, line 1370: So the problem here is that some systems still expect /sbin/dhcpcd to work without /usr available as it could be a network mount and it looks like libcrypto is found in /usr. See the tests using ALLOW_USR_LIBS elsewhere in configure.

tobhe: Right, makes sense. I guess putting the whole block behind `if [ -z "$OPENSSL" -a "$ALLOW_USR_LIB" = yes ]` would solve that problem.

configure
```diff
@@ -1349,6 +1352,53 @@ if [ "$OS" = dragonfly ]; then
 	echo "#else" >>$CONFIG_H
 fi
 
+if [ -z "$OPENSSL" ]; then
```
rsmarples, line 1355: We need to move this to below the libc and libmd checks for hmac and sha256 and only test for openssl if any of hmac or sha256 are not available.

For systems where we have lower level support, we should prefer these as they are generally availalable in /lib

compat/crypt/hmac.c
```diff
@@ -34,6 +34,35 @@
 
 #include "config.h"
 
+#ifdef WITH_OPENSSL
```
rsmarples, line 37: I would rather move this to it's own file, maybe openssl sub directory here or hmac_openssl.c
My rationale is that this is file is taken from upstream and would make syncing any changes harder. Not that I expect that to happen as the code is pretty mature :)

compat/crypt/sha256.c
```diff
@@ -40,6 +40,38 @@
 #include "config.h"
 #include "sha256.h"
 
+#ifdef WITH_OPENSSL
```
rsmarples, line 43: I would rather move this to it's own file, maybe openssl sub directory here or sha256_openssl.c
My rationale is that this is file is taken from upstream and would make syncing any changes harder. Not that I expect that to happen as the code is pretty mature :)

rsmarples (changes requested):

configure
tobhe: Would you mind if I move thing around a bit and put the hmac and sha checks before md5? If we fall back to openssl for hmac we don't need md5, so that would make things easier.

Another thing I noticed is that there is a dragonfly specific path where we already use `<openssl/sha.h>`. I wonder if that could be dropped with proper OpenSSL detection.

rsmarples: Sure, you can do that.

I can test any Dragonfly specific changes as it's currently my host OS for development - as it runs QEMU accelerated by NVMM better than NetBSD ;)

tobhe:
Pushed another update based on the latest feedback. The prioritization in `configure` is tricky to get right but I think the current version does a good job in doing what you requested. When anything other than openssl is available it will use that first, unless `--with-openssl` is explicitly passed as an argument.

I also moved the openssl shims to `compat/crypt_openssl` to leave the NetBSD files untouched.

I have tested on Ubuntu and OpenBSD, as mentioned above I would appreciate feedback on the DragonflyBSD diff. I left the workaround for MD5 in place but removed the "openssl/sha.h" part since that should now be handled automatically :)

rsmarples (approved):

configure
```diff
@@ -1521,7 +1491,95 @@ EOF
...
-if [ "$HMAC" = no ]; then
+
+if [ -z "$OPENSSL" ] && [ "$ALLOW_USR_LIBS" = true ] &&
+   [ "$SHA2" = no ] && [ "$HMAC" = no ]; then
+	printf "Testing for openssl ... "
+	cat <<EOF >_openssl.c
+#include <stdio.h>
+#include <openssl/opensslv.h>
+int main(void) {
+	return OPENSSL_VERSION_NUMBER;
```
rsmarples, line 1502: Why do we need this check? Surely the check for sha below is enough as that's what we need?

configure
```diff
@@ -1521,7 +1491,95 @@ EOF
...
+#include <stdio.h>
+#include <openssl/sha.h>
+
+int main(void) {
+	SHA256_CTX ctx;
+	SHA256_Init(&ctx);
+	return 0;
+}
+EOF
+	if $XCC _openssl_sha.c -lcrypto -o _openssl_sha 2>&3; then
```
rsmarples, line 1526: We'll need to use `$PKG_CONFIG` for CFLAGS and LDFLAGS here. See libudev below for an example.
For example, openssl could be installed into /usr/local or /usr/pkg.

configure
```diff
@@ -1521,7 +1491,95 @@ EOF
 	echo "$HMAC"
 	rm -f _hmac.c _hmac
 fi
-if [ "$HMAC" = no ]; then
+
+if [ -z "$OPENSSL" ] && [ "$ALLOW_USR_LIBS" = true ] &&
+   [ "$SHA2" = no ] && [ "$HMAC" = no ]; then
```
rsmarples, line 1496: This logic looks wrong. We want openssl if we don't have SHA or HMAC, not SHA and HMAC.

rsmarples (changes requested):

configure
tobhe: My idea was having one general is openssl available check and one specifically for the SHA api which is only available in some versions. If openssl is there we always use hmac from there, if SHA is available from openssl we use that, otherwise we use the `crypt_openssl/sha.c` EVP wrapper.

configure
tobhe: ack

configure
tobhe: The idea is to only check for OpenSSL if neither sha nor hmac are there, no? Or to put it differently: If either were detected we don't check for openssl.

tobhe: Oh, gotcha. If any of them is not available is what you said above, not if neither is available. Will fix it in the next version.

tobhe:
Newest iteration uses `pkg-config` to find libcrypto and hopefully fixed the `if else` jungle.

perkelix:
Does dhcpcd even need to ship its own implementations in compat? It would make more sense to always build against upstream libraries that see active maintenance.

rsmarples:
There are several OS that exist where nothing is used in compat as these functions exist within libc.

I don't even know if all these functions exist in any libraries available in Linux so it makes more sense to do it like this because as I said earlier some systems expect /usr to be network mounted after dhcpcd has started so just blindly using openssl isn't an option as it generally lives in /usr.

borkra:
A lot of the functions you have in compat are a part of libbsd in Linux, I am not sure making libbsd a requirement is a good thing, since it's not a required install.

perkelix:
Which would be this library: https://libbsd.freedesktop.org/

borkra:
Making OpenSSL a requirement, is not a good thing. Big distros like Ubuntu have all crypto libraries installed, but if you look on Embedded use, things are much more fragmented. There are OpenSSL, GnuTLS, wolfSSL and even "No SSL" builds... Are you going to support them all? OpenSSL is huge I wouldn't want to be required to include it.

perkelix:
This can currently build using the minimal libmd instead of the built-ins shipped in compat/crypto.

My guess is that Ubuntu would just rather use OpenSSL for all crypto. As long as this is configurable at build time and doesn't remove the possibility to use libmd or built-ins for other distros, we're good.

tobhe:
> Making OpenSSL a requirement, is not a good thing

I am not planning to limit the portability  of dhcpcd in any way or even change the default.

> My guess is that Ubuntu would just rather use OpenSSL for all crypto

Right, technically

borkra:
Looking at libbsd on Linux, in detail, it includes libmd with all SHA functions, maybe we just replace compat with libbsd on Linux.

tobhe:
Uploaded one more tiny fix to not have HAVE_MD5_H defined when using openssl.

I think the rest of the `compat` code is fine as it is mostly thin wrappers around existing libc api, some have even found their way into glibc like `arc4random`.
My plan is to make sure all of them are up to date with their respective *BSD versions and if not open a separate PR to sync them.

tobhe:
I am sure the config refactoring wasn't easy to follow, so here's what the current version should default to:
- If any of SHA2 or HMAC are available directly, use those.
- If not check if OpenSSL libcrypto is available and use it for hmac via `compat/crypt_openssl/hmac.c`.
  With OpenSSL we don't need MD5 because that is only used in `compat/hmac.c`.
  - If libcrypto implements SHA2 directly (older versions and LibreSSL) we use that
  - Otherwise we use the new `compat/crypt_openssl/sha2.c` which implments the SHA2 interface via the higher level `EVP_*` API.
- If neither SHA/HMAC nor OpenSSL are available use `compat/crypt/*`

tobhe:
Any advice on how we could move forward with this? Would it make more sense to not change the defaults for now and only enable openssl support if `--with-openssl` is explicitly passed as an argument?

perkelix:
@rsmarples have we reached a conclusion about this (and other pending) merge requests?

perkelix:
@ido Any feedback?

ido:
> @ido Any feedback?

@perkelix will take a look this weekend.

tobhe:
One more update. I added an explicit call to `OPENSSL_init_crypto()` early during daemon startup because some versions of openssl do this lazily on first use of the API which clashes with seccomp. I also updated the `configure` test to probe for the same API.
Since this is a one-time setup action this seems like the better way rather than explicitly allowing the syscalls needed for this later.

rsmarples:
That's always the danger of linking which I try and avoid in the first place :)

I'm on holiday next week and without internet and will try and look at this some more when I get back.

perkelix:
> I'm on holiday next week and without internet and will try and look at this some more when I get back.

@rsmarples, is there any progress on this?

rsmarples:
@perkelix sorry for the late reply, my personal life has been interesting of late.

@tobhe I've been hammering this PR on a lot of various OS's can cannot find any issue with it, so from the technical perspective it's good to merge and well done for getting here!

I'll spend some time having a final look over the weekend to refresh and then merge it.

PR #226, opened 2023-07-17
compat sync

tobhe:
Here is a set of small compat syncs. None of them should make a functional difference, most actually just adjust comments.

I was reviewing them and checking that they aren't missing any fixes so I thought I might as well just sync them while I'm there.

One thing missing here is arc4random.c which seems to have diverged quite a bit from all shipped versions. Would there be interest in having that synced to an upstream version too or is there a reason for the divergence?


rsmarples:
arc4random has some changes by me to support privsep, fixing some compiler warnings and sweeping some coverity (which I don't really use anymore) under the table.

I'm happy to review any sync changes you want to bring in though.

Thanks for your work so far!

PR #227, opened 2023-07-20
compat: update arc4random() to newer chacha20 based version from OpenBSD

tobhe:
This updates the compat version to the modern chacha20 based arc4random implementation shipped in most systems nowadays. I merged upstream's arc4random.h into arc4random.c to make things a bit simpler.

There certainly is room for further improvement but I don't want to make this diff even bigger.
The fallback entropy mechanism if `/dev/urandom` doesn't work is very weak, a more sophisticated alternative could borrow parts from the one shipped in libressl https://github.com/openbsd/src/blob/master/lib/libcrypto/arc4random/getentropy_netbsd.c
If OpenSSL is available one could also use their PRNG as entropy source.


compat/arc4random.c
```diff
@@ -1,173 +1,290 @@
...
+	int		 fd;
+	uint8_t		*rand = (uint8_t *)buf;
+
+	if (length < sizeof(tv)) {
+		gettimeofday(&tv, NULL);
+		memcpy(buf, &tv, sizeof(tv));
+		length -= sizeof(tv);
+		rand += sizeof(tv);
+	}
+	fd = open("/dev/urandom", O_RDONLY | O_NONBLOCK);
```
rsmarples, line 101: This won't work in an empty chroot which is part of our privsep model.
We need to ensure that the fd is kept open and re-used which is the main change I did to the prior arc4random code.

tobhe:
ok, so I finally got around to do some testing and found that the previous approach of storing fd in `struct rs` was not the best decision because of the order in which they are used/initialized.
I have moved the fd to a separate static variable instead which isn't nice but works better.

tobhe:
I have run some tests over the last few hours. For that I have added `loginfox()` for each reseed and each `arc4random()` call, set `REKEY_BASE` to 64 to force more frequent stir operations, and explicitly passed `--without-arc4random --without-arc4random-uniform`.
Everything looks good as far as I can tell. Below is some example log output:
```
Aug 16 12:00:13 [3797]: arc4random: 60a28118 = arc4random()
Aug 16 12:01:08 [3797]: arc4random: a86e7023 = arc4random()
Aug 16 12:02:04 [3797]: _dhcpcd_getentropy: read() = 43 81 8b f3 70 b5 55 f4 9f c6 46 7c 02 d5 d1 ba 58 5a b4 d8 a8 92 ed df f0 a5 5c fc c1 66 d9 49 f0 0d 43 55 b3 c5 87 89
Aug 16 12:02:04 [3797]: arc4random: d421a5eb = arc4random()
Aug 16 12:02:57 [3797]: arc4random: 00baa982 = arc4random()
Aug 16 12:03:53 [3797]: arc4random: 2fb49bcc = arc4random()
Aug 16 12:04:49 [3797]: arc4random: 7f8f3c95 = arc4random()
Aug 16 12:05:44 [3797]: arc4random: 804715ff = arc4random()
Aug 16 12:06:37 [3797]: arc4random: 987585df = arc4random()
Aug 16 12:07:33 [3797]: arc4random: d6cf2df1 = arc4random()
Aug 16 12:08:30 [3797]: arc4random: 3ec58e59 = arc4random()
Aug 16 12:09:26 [3797]: arc4random: cc30d744 = arc4random()
Aug 16 12:10:21 [3797]: arc4random: 56cf71c4 = arc4random()
Aug 16 12:11:17 [3797]: arc4random: e41760fe = arc4random()
Aug 16 12:12:13 [3797]: arc4random: 314965e0 = arc4random()
Aug 16 12:13:07 [3797]: arc4random: 51d2211b = arc4random()
Aug 16 12:14:02 [3797]: arc4random: 4d4a8d64 = arc4random()
Aug 16 12:14:56 [3797]: arc4random: 30404183 = arc4random()
Aug 16 12:15:50 [3797]: arc4random: 2b159c7a = arc4random()
Aug 16 12:16:45 [3797]: arc4random: c5469dcd = arc4random()
Aug 16 12:17:38 [3797]: arc4random: 543ef661 = arc4random()
Aug 16 12:18:34 [3797]: arc4random: df561bef = arc4random()
Aug 16 12:19:29 [3797]: arc4random: 941c825e = arc4random()
Aug 16 12:20:25 [3797]: _dhcpcd_getentropy: read() = 7f ec 23 35 76 3e f7 31 ad c2 18 0d c6 be 8f 46 74 c6 28 48 29 f7 3b c1 f3 04 da 2b 5e 23 32 f7 82 bd fb c6 62 b5 41 18
Aug 16 12:20:25 [3797]: arc4random: 6237cd08 = arc4random()
Aug 16 12:21:20 [3797]: arc4random: 047d2ae6 = arc4random()
Aug 16 12:22:14 [3797]: arc4random: ffbe312a = arc4random()
Aug 16 12:23:09 [3797]: arc4random: 284d05c5 = arc4random()
```
The `read()` tells us it actually uses the fd, the output it gets also looks random and different on each call.

Technically speaking with newer PRNGs like `arc4random` one nice property is that there is no such thing as diluting the entropy pool. Seeding in bad entropy does not make the output worse, which has been a problem with older algorithms. Since we always seed at least once from a good entropy source before switching to a `chroot`, even if the follow up stirs use a worse entropy source, the impact wouldn't be too bad. Using a good source if one is available still seems better of course.

PR #229, opened 2023-07-22
Add automated CI builds for Ubuntu, OpenBSD, FreeBSD and NetBSD

tobhe:
I have added some build automation to my local branch to help me catch eventual bugs in my commits and I thought this might be something worth adding here.

With the proposed configuration, automated builds are triggered for each commit to the master branch and for each pull request against the master branch. I guess this would be quite helpful to find out if a PR breaks any of the supported systems (or at least those we can run in a CI) without testing them all by hand.

The BSDs are not officially supported by github, but I have successfully used https://github.com/mario-campos/emulate for another portable daemon I am working on. It works by creating and running BSD VMs in macos. The builds take a little longer than native ones but otherwise they work just fine.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,86 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  CC: clang
+
+jobs:
+  ubuntu:
+    strategy:
+      matrix:
+        os: [ ubuntu-latest, ubuntu-22.04 ]
+    runs-on: ${{ matrix.os }}
+
+    steps:
+    - uses: actions/checkout@v2
+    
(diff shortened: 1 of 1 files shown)


tobhe:
An example of how this looks like is available here: https://github.com/tobhe/dhcpcd/pull/2
The build results can be viewed on the bottom next to the green checkmark by clicking `Show all checks`. If any of them fail it turns to a red cross and shows a warning.

perkelix:
Would the Ubuntu one work on Debian as-is?

igalic:
except for alpine ;)

rsmarples:
In the dubious free time I have right now, I did look into using my BSD dev machine as build host for this but this might be better for the here and now.

rsmarples:
> except for alpine ;)

I do test on alpine just for musl

PR #236, opened 2023-08-09
dhcpcd: Fix off-by-one overflow when read() writes full BUFSIZ

tobhe:
Read can return up to `BUFSIZ` bytes. If that happens we zero `log[BUFSIZ]` which is outside the buffer. Found with a long error message and `-fsanitize=address`.

### DIFF
--- a/src/dhcpcd.c
+++ b/src/dhcpcd.c
@@ -1822,7 +1822,7 @@ dhcpcd_stderr_cb(void *arg, unsigned short events)
 	if (!(events & ELE_READ))
 		return;
 
-	len = read(ctx->stderr_fd, log, sizeof(log));
+	len = read(ctx->stderr_fd, log, sizeof(log) - 1);
 	if (len == -1) {
 		if (errno != ECONNRESET)
 			logerr(__func__);


src/dhcpcd.c
```diff
@@ -1829,7 +1829,7 @@ dhcpcd_stderr_cb(void *arg, unsigned short events)
 		return;
 	}
 
-	log[len] = '\0';
+	log[len < BUFSIZ ? len : BUFSIZ - 1] = '\0';
```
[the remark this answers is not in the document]
tobhe: But passing `sizeof(log) - 1` doesn't actually null terminate the buffer unless we also initialize it to all zeros right? `s/BUFSIZ/sizeof(log)/` sounds good.

tobhe:
Does that look better to you?

src/dhcpcd.c
```diff
@@ -1822,14 +1822,14 @@ dhcpcd_stderr_cb(void *arg, unsigned short events)
...
-	len = read(ctx->stderr_fd, log, sizeof(log));
+	memset(log, 0, sizeof(log));
+	len = read(ctx->stderr_fd, log, sizeof(log) - 1);
 	if (len == -1) {
 		if (errno != ECONNRESET)
 			logerr(__func__);
 		return;
 	}
 
-	log[len] = '\0';
```
rsmarples, line 1832: Please restore this line - we should explicitly terminate still as we're dumping data from a fd to stderr.

rsmarples:
> Does that look better to you?

Yes, just one minor comment.
Out of curiosity, what data did you have to trigger this?

tobhe:
>  Out of curiosity, what data did you have to trigger this?

I was debugging a crash caused by the OpenSSL symbol conflict with `--fsanitize=address` and hit this. I think it tried to print a very long stacktrace.

src/dhcpcd.c
tobhe: done

src/dhcpcd.c
```diff
@@ -1822,7 +1822,8 @@ dhcpcd_stderr_cb(void *arg, unsigned short events)
 	if (!(events & ELE_READ))
 		return;
 
-	len = read(ctx->stderr_fd, log, sizeof(log));
+	memset(log, 0, sizeof(log));
```
rsmarples, line 1825: Excellent! Now this memset isn't needed as we read into the buffer and explicitly terminate.

tobhe: urgh right. fixed.

rsmarples:
Thanks!

PR #248, opened 2023-09-25
compat: use OpenSSL RAND_priv_bytes() for entropy

tobhe:
With the recent OpenSSL addition we can now also use it in a few other places for convenience.

One useful thing it implements is portable high quality entropy generation. We can use it as a replacement for `/dev/urandom` to seed `arc4random()` where it is available.

### DIFF
--- a/compat/arc4random.c
+++ b/compat/arc4random.c
@@ -40,6 +40,10 @@
 #include <sys/types.h>
 #include <sys/time.h>
 
+#if defined(HAVE_OPENSSL)
+#include <openssl/rand.h>
+#endif
+
 #define KEYSTREAM_ONLY
 #include "chacha_private.h"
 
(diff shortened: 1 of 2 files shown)


rsmarples:
Doesn't OpenSSL itself seed from `/dev/urandom`? I'm not entirely sure what the gain is here, to me it just looks like using OpenSSL because we can rather than anything else?

Tests on Ubuntu are now failing.

tobhe:
> Doesn't OpenSSL itself seed from /dev/urandom? I'm not entirely sure what the gain is here, to me it just looks like using OpenSSL because we can rather than anything else?

It actually seeds from multiple sources and can even be extended with providers to use additional sources. See https://www.openssl.org/docs/manmaster/man7/EVP_RAND.html and the default implementations in https://github.com/openssl/openssl/tree/master/providers/implementations/rands/seeding. It should even work in in cases where there is no urandom available.

Another benefit for us is that we can do additional entropy checks and similar things required for certifications in OpenSSL without needing to modify dhcpcd itself.

> Tests on Ubuntu are now failing.

Also fixed in newest upload

Issue #246, opened 2023-09-20
Prefix Exclude Option is malformed in DHCPv6-Request


ctomahogh:
I am using dhcpcd-10.0.2 to obtain an IP address and a Prefix from a DHCPv6 server.

The server is not under my control, so I cannot change the config or test other configurations.
My dhcpcd config file looks roughly like this

````
quiet
script /bin/dhcpc_helper
nogateway
nodev
noipv4ll
option domain_name_servers

hostname
noipv4
nodhcp
ipv6
dhcp6
interface eth0
  ia_na 0
  ia_pd 1 eth1/2 eth3/3 
````
In the advertise message the server sends not only a Prefix delegation, but also adds a Prefix exclude option.
The provided prefix is i.e. 2001:db8:12:abc0:: /62
And the exclude option excludes the prefix 2001:db8:12:abc0:: /64

So the Prefix Exclude Option in the Advertise message has a length of 2 and the first byte of the option payload is 64 (prefix length) and the second byte (IPv6 subnet ID) is 0

![advertise_dp_with_pe](https://github.com/NetworkConfiguration/dhcpcd/assets/22915042/61b6d25c-db59-4da1-aa9d-5f026e2adfcd)

Now the dhcpcd sends a DHCPv6-Request but with a malformed Prefix Exclude Option. The length of the Prefix Exclude option is 0

![request_dp_with_malformed_pe](https://github.com/NetworkConfiguration/dhcpcd/assets/22915042/f8dc06b2-3b1a-453f-8f1a-3e66ed35c45f)

Obviously the DHCPv6 server I am using does not like this and therefor does not send a  Reply.

As far as I understand RFC 6603 this is not correct. Actually the Request message should in this case reply with a length of 2 and the prefix length 64 and IPv6 subnet ID 0 (like in the advertise)

I looked into the code in dhcp6.c in the function `dhcp6_makemessage` where the request message is created. In my opinion the according source code (currently starting from line 1007 in master branch) is buggy and **always** sends a prefix exclude option with length 0

```
/* RFC6603 Section 4.2 */
if (ap->prefix_exclude_len) {
    uint8_t exb[16], *ep, u8;
    const uint8_t *pp;

    n = (size_t)((ap->prefix_exclude_len -
	ap->prefix_len - 1) / NBBY) + 1;
    ep = exb;
    *ep++ = (uint8_t)ap->prefix_exclude_len;
    pp = ap->prefix_exclude.s6_addr;
[...]

PR #250, opened 2023-09-28
Send correct amount of used buffer for prefix exclude option

ctomahogh:
Fixes issue #246 

The payload of the prefix exclude option was correctly created but the amount of bytes to send in the DHCPv6 request was always set to 0 which resulted in an invalid prefix exclude option

This patch fixes this behavior by calculating the correct amount of bytes to send.


rsmarples:
Thanks!

PR #252, opened 2023-10-08
compat/arc4random.c: use memset instead of explicit_bzero

ffontaine:
Use `memset` instead of `explicit_bzero` to avoid the following build failure with uclibc-ng since version 10.0.3 and
https://github.com/NetworkConfiguration/dhcpcd/commit/837d09e34c487edaa92aa2ae71a630d84c927f8e:

```
/home/fabrice/buildroot/output/host/lib/gcc/arm-buildroot-linux-uclibcgnueabi/12.3.0/../../../../arm-buildroot-linux-uclibcgnueabi/bin/ld: ../compat/arc4random.o: in function `_rs_stir_if_needed':
arc4random.c:(.text+0x8cc): undefined reference to `explicit_bzero'
```

### DIFF
--- a/compat/arc4random.c
+++ b/compat/arc4random.c
@@ -195,7 +195,7 @@ _rs_stir(void)
 		_rs_init(rnd, sizeof(rnd));
 	else
 		_rs_rekey(rnd, sizeof(rnd));
-	explicit_bzero(rnd, sizeof(rnd));	/* discard source seed */
+	memset(rnd, 0, sizeof(rnd));	/* discard source seed */
 
 	/* invalidate rs_buf */
 	rs->rs_have = 0;


igalic:
C23 gives us `memset_explicit()`, so maybe in 3-10 years we can start using it

thithib:
With this change, chances are the call to `memset()` will be optimized out by the compiler and the `rnd` buffer won't be erased. I don't know the security impact of that but I assume there was a good reason to use `explicit_bzero()` in the first place.

Wouldn't it be safer to leverage feature test macros or the like to keep relying on `explicit_bzero()` on systems that support it and use alternatives on others?

rsmarples:
Happy to look at a PR that adds `HAVE_MEMSET_EXPLICT` and `HAVE_BZERO_EXPLICIT` macros set by configure and call the relevant functions accordingly. If anyone is concerned by this in dhcpcd-10.0.4 they can always set `CFLAGS=-O0` which should disable any optimisations.

rsmarples:
@thithib This is now done and I emitted a warning if we have to fallback to memset.
Only uclibc users (which @ffontaine is using based on his output) will see this warning hopefully.

PR #254, opened 2023-10-16
privsep: allow __NR_clock_gettime32 syscall

olegartys:
musl libc doesn't have __NR_clock_gettime definition, but has __NR_clock_gettime32. clock_gettime implementation fallbacks to 32-bit version if 64-bit is not supported by the kernel.


rsmarples:
Thanks!

Issue #282, opened 2024-01-15
[feature request] dump unknown options using the `--dumplease` flag


holmanb:
Unknown lease options don't appear to be dumped by dhcpcd. Microsoft Azure, for example, uses dhcp option 245 to announce its wireserver IP address to cloud instances. The ability to use `dhcpcd --dumplease` would be very helpful for use cases like this.

```bash
$ dhcpcd --dumplease < /var/lib/dhcpcd/eth0.lease 
broadcast_address='10.0.0.255'
classless_static_routes='0.0.0.0/0 10.0.0.1 168.63.129.16/32 10.0.0.1 169.254.169.254/32 10.0.0.1'
dhcp_lease_time='4294967295'
dhcp_message_type='5'
dhcp_rebinding_time='4294967295'
dhcp_renewal_time='4294967295'
dhcp_server_identifier='168.63.129.16'
domain_name='ilo2tr0xng2exgucxg20yx0tjb.gx.internal.cloudapp.net'
domain_name_servers='168.63.129.16'
ip_address='10.0.0.5'
network_number='10.0.0.0'
routers='10.0.0.1'
server_name='DSM111070915004'
subnet_cidr='24'
subnet_mask='255.255.255.0'
```

But I wrote some Python which demonstrates that this option is included in the lease:
```python
#!/usr/bin/env python3
import sys
import os

def iter_options(data: bytes, index: int):
    """options are variable length, and consist of the following format

    option number: 1 byte
    option length: 1 byte
    option data: variable length (see length field)
    """
    while len(data) >= index + 2:
        code = data[index]
        length = data[1 + index]
        option = data[2 + index : 2 + index + length]
        yield code, option
        index = 2 + length + index

lease_file = sys.argv[1]
print(f"parsing lease: {lease_file}")
with open(lease_file, "rb") as f:
    for code, option in iter_options(f.read(), 240):
        print(f" found option {code}: {option}")
```

Which shows that option 245 is in the lease:
```bash
$ ./get_options.py /var/lib/dhcpcd/eth0.lease 
parsing lease: /var/lib/dhcpcd/eth0.lease
 found option 1: b'\xff\xff\xff\x00'
 found option 3: b'\n\x00\x00\x01'
 found option 6: b'\xa8?\x81\x10'
 found option 15: b'ilo2tr0xng2exgucxg20yx0tjb.gx.internal.cloudapp.net'
 found option 51: b'\xff\xff\xff\xff'
[...]

PR #284, opened 2024-01-17
Add definition for option 245

lparkes:
Add definition for the DHCP private option used to advertise the location of the Azure WireServer.
See https://cloudinit.readthedocs.io/en/19.4/topics/datasources/azure.html.

I didn't add the definition to the small configuration it really isn't useful there.

### DIFF
--- a/src/dhcpcd-definitions.conf
+++ b/src/dhcpcd-definitions.conf
@@ -335,6 +335,10 @@ encap 255	flag			global
 
 # Options 224-254 are reserved for Private Use
 
+# Option 245 is an IANA assigned private number used by Azure DHCP servers
+# to provide the IPv4 address of the Azure WireServer endpoint
+define 245	ipaddress		azureendpoint
+
 # Option 249 is an IANA assigned private number used by Windows DHCP servers
 # to provide the exact same information as option 121, classless static routes
 define 249	rfc3442			ms_classless_static_routes


holmanb:
@lparkes Nice, thanks! This is all I really need for https://github.com/NetworkConfiguration/dhcpcd/issues/282, I'd be happy to close that one if this merges.

rsmarples:
@lparkes I cannot find any public reference for IANA or RFC DHCP option 245.
https://www.iana.org/assignments/bootp-dhcp-parameters/bootp-dhcp-parameters.xhtml

I don't argue your change is good, but I can't find any standard outside of Azure for it.

lparkes:
@rsmarples I believe you are correct. My understanding is that this is an Azure specific option in the IANA "Reserved (Private Use)" option range 224-254. For the comment in the commit I copied the phrasing of the similar option 249 (also in that same range).

Should a link/reference to some suitable Azure documentation be added to the comment in the config file?

rsmarples:
> @rsmarples I believe you are correct. My understanding is that this is an Azure specific option in the IANA "Reserved (Private Use)" option range 224-254. For the comment in the commit I copied the phrasing of the similar option 249 (also in that same range).
> 
> Should a link/reference to some suitable Azure documentation be added to the comment in the config file?

At a minimum I think yes. We strive to be a public service and this is very much a private option.
My only fear is what happens if IANA decide to give this option another meaning?

Currently dhcpcd only has RFC options (although a few have expired).

lparkes:
Your statement that "dhcpcd only has RFC options" is not entirely correct because it has a definition for option 249, which while the contents of that option have an RFC number, the option number itself is not assigned by an RFC or IANA. 

I am not concerned about option 245 being redefined for something else in the future because there are still plenty of unassigned option numbers for IANA to allocate. On top of that, if the site-specific "Reserved (Private Use)" range was redefined again (as in RFC 3942) then Microsoft could simply ask IANA to define option 245 to its Azure usage. RFC 3942 defined a process for this.

What I am concerned about is if someone decides to use option 245 in their own cloud for a similar service, but with a different definition. Maybe an IPv6 address instead of an IPv4 address. This is a risk for the entire range 224-254, including the currently implemented option 249.

We must support site-specific options in some way or other because the sites will set them in the DHCP servers and the clients need to get them somehow. If we are wary of decoding the site-specific options, then maybe put in 31 site specific binhex definitions. I don't actually know what a binhex definition is, but I think I can guess.

holmanb:
> If we are wary of decoding the site-specific options, then maybe put in 31 site specific binhex definitions. I don't actually know what a binhex definition is, but I think I can guess.

@lparkes I assume you're suggesting something like this?
```
# Options 224-254 are reserved for Private Use
define 224	binhex			private_option_224
define 225	binhex			private_option_225
define 226	binhex			private_option_226
...
define 254	binhex			private_option_254
```

That seems reasonable. Not assuming the length or type of the content of the option should be general purpose enough to meet the requirements of multiple sites. I assume that option 249 will need to retain its rfc3442 format for existing users?

lparkes:
Exactly. I would use the name "site_specific" instead of "private_option" because that is a phrase used in the RFCs and it also indicates to whom the option is private. I should be able to find the time to create an alternative draft PR in a day or two.

rsmarples:
I've been mulling this over for some time. I think I would like to merge both changes. Obviously they conflict, so you'll need to make some adjustments.

I think we need to add to the private use comment that the options we define such as this:
`These options are designed to be configured by the end user if needed and any configuration here may change in the future.`
A kind of disclaimer if you will. I'll leave the exact wording up to you.

I also think it needs be in this order

1) Commentary
2) Configured options such as 245 and 249
3) The options without any specific configuration - we need to treat it as binary data and expose it in hex - binhex.

@lparkes does this make sense?

PR #293, opened 2024-01-27
Example alternate option 245 implementation

lparkes:
Here is an example change that can be used if my continuing the example of option 249 is undesirable.

Added binhex entries for all the site-specific options to src/dhcpcd-definitions.conf so that site specific hooks can make use of them.

The entries were not added to the small definitions file because that seems neither useful nor small.

### DIFF
--- a/src/dhcpcd-definitions.conf
+++ b/src/dhcpcd-definitions.conf
@@ -333,16 +333,49 @@ encap 255	flag			global
 
 # Options 222 and 223 are unused, RFC3942
 
-# Options 224-254 are reserved for Private Use
+# Options 224-254 are reserved for Private Use, RFC3942
+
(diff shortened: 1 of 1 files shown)


holmanb:
Thanks @lparkes, this change looks good to me. If Roy is on board with this approach I'll happily test it.

perkelix:
It might help if someone actually tags @rsmarples if a comment or review is expected.

rsmarples:
> It might help if someone actually tags @rsmarples if a comment or review is expected.

No need to tag me unless I take too long to reply :)
Lloyd is going to merge this and the other PR into one I hope with the comments I made in #284

lparkes:
Closed in deference to PR #299.

Issue #102, opened 2022-05-14
IPv6 RA messages specific routes


cfilipem:
Is it possible to have dhcpcd handling IPv6 more specific routes received in IPv6 Router Advertisement messages, as described in [RFC4191](https://www.rfc-editor.org/rfc/rfc4191.html)?

goertzenator:
A motivating use case:

1. You have a router which provides internet access (GUA prefix) and offers a default route.
2. You add a separate VPN router (ULA prefix) onto the same link which is supposed to handle all traffic on your corporate extranet.  This router advertises a non-default route for a /48 version of the ULA prefix, but no default route.

At the moment dhcpcd will not pick up the non-default route from the VPN router and will just push everything down the default route.

goertzenator:
I am taking a swing at this now.

goertzenator:
Basic functionality is working, but I need to test more and read RFC4191 more thoroughly to see if there is stuff I should add.

https://github.com/goertzenator/dhcpcd/tree/rfc4191route

perkelix:
Does the above still apply to 10.5.2?

goertzenator:
A fix was merged a while ago.  This issue should be closed.

PR #297, opened 2024-02-21
add RFC4191 support

goertzenator:
Resolves issue #102.

Implement RFC4191 support for Route Information options in Router Advertisements.  This option allows a router to provide routes for a specific prefix instead of only providing default routes.

My use case is that I have a ULA network with a VPN router and no IPv6 internet access.  Using this router as the default route does not work well because the host then thinks it can reach the Internet.  By instead advertising a route just for the ULA prefix hosts will immediately know they cannot reach the Internet through this router.

I've tested this PR in a VM in this exact configuration and it seems to work quite well.  I changed the radvd config on the fly to remove and add the route and it behaved as expected.

### DIFF
--- a/src/ipv6nd.c
+++ b/src/ipv6nd.c
@@ -71,6 +71,20 @@
 #define	ND_OPT_PI_FLAG_ROUTER	0x20	/* Router flag in PI */
 #endif
 
+#ifndef ND_OPT_RI
+#define ND_OPT_RI	24
+struct nd_opt_ri {		/* Route Information option RFC4191 */
+	uint8_t	 nd_opt_ri_type;
+	uint8_t	 nd_opt_ri_len;
+	uint8_t	 nd_opt_ri_prefixlen;
+	uint8_t	 nd_opt_ri_flags_reserved;
+	uint32_t nd_opt_ri_lifetime;
+	struct in6_addr nd_opt_ri_prefix;
+};
+__CTASSERT(sizeof(struct nd_opt_ri) == 24);
+#define OPT_RI_FLAG_PREFERENCE(flags) ((flags & 0x18) >> 3)
+#endif
+
 #ifndef ND_OPT_RDNSS
 #define ND_OPT_RDNSS			25
 struct nd_opt_rdnss {           /* RDNSS option RFC 6106 */
@@ -612,10 +626,10 @@ ipv6nd_startexpire(struct interface *ifp)
 }
 
 int
-ipv6nd_rtpref(struct ra *rap)
+ipv6nd_rtpref(uint8_t flags)
 {
 
-	switch (rap->flags & ND_RA_FLAG_RTPREF_MASK) {
+	switch (flags & ND_RA_FLAG_RTPREF_MASK) {
 	case ND_RA_FLAG_RTPREF_HIGH:
 		return RTPREF_HIGH;
 	case ND_RA_FLAG_RTPREF_MEDIUM:
@@ -624,7 +638,7 @@ ipv6nd_rtpref(struct ra *rap)
 	case ND_RA_FLAG_RTPREF_LOW:
 		return RTPREF_LOW;
 	default:
-		logerrx("%s: impossible RA flag %x", __func__, rap->flags);
+		logerrx("%s: impossible RA flag %x", __func__, flags);
 		return RTPREF_INVALID;
 	}
 	/* NOTREACHED */
@@ -649,7 +663,7 @@ ipv6nd_sortrouters(struct dhcpcd_ctx *ctx)
 				continue;
 			if (!ra1->isreachable && ra2->reachable)
 				continue;
-			if (ipv6nd_rtpref(ra1) <= ipv6nd_rtpref(ra2))
+			if (ipv6nd_rtpref(ra1->flags) <= ipv6nd_rtpref(ra2->flags))
 				continue;
 			/* All things being equal, prefer older routers. */
 			/* We don't need to check time, becase newer
@@ -827,6 +841,7 @@ ipv6nd_removefreedrop_ra(struct ra *rap, int remove_ra, int drop_ra)
 	if (remove_ra)
 		TAILQ_REMOVE(rap->iface->ctx->ra_routers, rap, next);
 	ipv6_freedrop_addrs(&rap->addrs, drop_ra, NULL);
+	routeinfohead_free(&rap->routeinfos);
 	free(rap->data);
 	free(rap);
 }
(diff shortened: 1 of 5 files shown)


src/ipv6nd.c
```diff
@@ -2135,3 +2214,53 @@ ipv6nd_startrs(struct interface *ifp)
...
+
+	ri = calloc(1, sizeof(struct routeinfo));
+	if (ri == NULL)
+		return NULL;
+
+	memcpy(&ri->prefix, prefix, sizeof(ri->prefix));
+	ri->prefixlength = prefixlength;
+	ri->flags = flags;
+	ri->lifetime = lifetime;
+	clock_gettime(CLOCK_MONOTONIC, &ri->acquired);
```
rsmarples, line 2232: We could use the  acquired time of the RA here instead of a call to get the monotonic clock.

src/sa.c
```diff
@@ -300,11 +300,39 @@ sa_toprefix(const struct sockaddr *sa)
 	return prefix;
 }
 
+void
+ipbytes_fromprefix(uint8_t *ap, int prefix, int max_prefix)
```
rsmarples, line 304: Why did you change this? It's not used outside of sa in this branch AFAIK.

src/ipv6nd.h
```diff
@@ -105,7 +120,7 @@ int ipv6nd_open(bool);
 int ipv6nd_openif(struct interface *);
 #endif
 void ipv6nd_recvmsg(struct dhcpcd_ctx *, struct msghdr *);
-int ipv6nd_rtpref(struct ra *);
+int ipv6nd_rtpref(uint8_t);
```
rsmarples, line 123: Why did you change this?

src/ipv6nd.c
```diff
@@ -2135,3 +2214,53 @@ ipv6nd_startrs(struct interface *ifp)
...
 	return;
 }
+
+struct routeinfo *routeinfo_new(const struct in6_addr *prefix, uint8_t prefixlength, uint8_t flags, uint32_t lifetime)
+{
+	struct routeinfo *ri;
+	char buf[INET6_ADDRSTRLEN];
+	const char *p;
+
+	ri = calloc(1, sizeof(struct routeinfo));
```
rsmarples, line 2224: malloc should be sufficient here.
Yes there will be undefined bytes inside the prefix as length will likely be less, but that's fine as we always access limited via prefixlength.

src/ipv6nd.h
```diff
@@ -37,6 +37,20 @@
...
 #include "ipv6.h"
 
+/* rfc4191 */
+struct routeinfo {
+	TAILQ_ENTRY(routeinfo) next;
+	struct in6_addr prefix;
+	uint32_t lifetime;
+	struct timespec acquired;
+	uint8_t flags;
+	uint8_t prefixlength;
```
rsmarples, line 47: please rename to prefix_len and move to underneath the prefix to match ipv6.h
Move acquired to just above sprefix.

src/ipv6nd.h
```diff
@@ -45,13 +59,14 @@ struct ra {
...
 	struct timespec acquired;
-	unsigned char flags;
+	uint8_t flags;
 	uint32_t lifetime;
 	uint32_t reachable;
 	uint32_t retrans;
 	uint32_t mtu;
 	uint8_t hoplimit;
 	struct ipv6_addrhead addrs;
+	struct routeinfohead routeinfos;
```
rsmarples, line 69: Please rename to rinfos.
To save many comments, please rename all local routeinfo variables to rinfo.

rsmarples (changes requested):
Technically it looks sound.
I've made some comments mainly on style but some recommendations also.

Thanks for the effort thus far!

src/ipv6nd.h
goertzenator: Would you like `struct routeinfo` to `struct rinfo` too? (same for `routeinfohead` -> `rinfohead`)

rsmarples: No routeinfo is fine for the struct name.

src/sa.c
goertzenator: We compute a netmask from a prefix_len [here](https://github.com/goertzenator/dhcpcd/blob/rfc4191route/src/ipv6.c#L2339).  The original fromprefix function only operated on a `struct sockaddr` which `rinfo` does not have or need.  I unwrapped the prefix code in sa.c to make it usable for `rinfo`.

`ipbytes_fromprefix()` should be static and removed from sa.h since it is indeed only used inside sa.c.

src/ipv6nd.c
```diff
@@ -1502,6 +1520,51 @@ ipv6nd_handlera(struct dhcpcd_ctx *ctx,
...
+			memcpy(&ri, p, olen); /* may be smaller than sizeof(ri), pad with zero */
+			if(ri.nd_opt_ri_prefixlen > 128) {
+				logmessage(loglevel, "%s: invalid route info prefix length",
+				    ifp->name);
+				break;
+			}
+
+			/* rfc4191 3.1 - RI for ::/0 applies to default route */
+			if(ri.nd_opt_ri_prefixlen == 0) {
+				rap->lifetime = ntohl(ri.nd_opt_ri_lifetime);
```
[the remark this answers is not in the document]
rsmarples: The warning here:
https://github.com/goertzenator/dhcpcd/blob/rfc4191route/src/ipv6nd.c#L1258

Needs to be moved to here:
https://github.com/goertzenator/dhcpcd/blob/rfc4191route/src/ipv6nd.c#L1602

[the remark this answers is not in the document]
goertzenator: Ah, I understand now.  Good catch!

goertzenator:
Thanks for the prompt review, I appreciate it!

src/ipv6nd.c
```diff
@@ -1502,6 +1520,52 @@ ipv6nd_handlera(struct dhcpcd_ctx *ctx,
...
+				rap->flags = ((rap->flags & (~ (unsigned int)ND_RA_FLAG_RTPREF_MASK))
+					| ri.nd_opt_ri_flags_reserved) & 0xff;
+
+				break;
+			}
+
+			/* Update existing route info instead of rebuilding upon RA so that
+			previously announced but now absent routes can stay alive.  To kill a
+			route early, an RI with lifetime=0 needs to be received (rfc4191 3.1)*/
+			rinfo = routeinfo_find(rap, &ri.nd_opt_ri_prefix, ri.nd_opt_ri_prefixlen);
```
rsmarples, line 1551: I would make this simpler.

Rename routeinfo_find to routeinfo_alloc (or findalloc) and pass the same arguments as here.
Then below if NULL, break (ensure error reported somewhere) if not then assign the rest of the data.
This avoids two different places to assign the same data.

goertzenator: Changes made as requested.

rsmarples:
Thanks for your contribution!

PR #299, opened 2024-02-25
Define the Azure Endpoint and other site-specific options

lparkes:
Added the azureendpoint site-specific option as an ipaddress definition to make it easier for Azure VMs using dhcpcd to get their WireServer endpoint address.

Added binhex definitions for all otherwise undefined site-specific options so that site-specific hooks can use them.


lparkes:
This PR supersedes PR #284 and PR #293.

rsmarples:
Thanks!

PR #307, opened 2024-03-12
dhcpcd: Respect IPV6_PREFERRED_ONLY flag regardless of state

taoyl-g:
Current IPv6_PREFERRED_ONLY (option 108) handling code is only effective when current state is DHS_DISCOVER or DHS_REBOOT. However the assumption that we'll only receive OFFER and ACK with option 108 in those two states is not true. Particularly, when we receive multiple ACKs upon our REQUEST, the first ACK will trigger the use_v6only code path and dhcp_drop() us into DHS_NONE state, therefore the option 108 on second ACK won't be handled correctly and we'll bind to the lease incorrectly.

DHS_NONE is not the only possible state, considering the corner cases of server configuration change and multiple servers. Thus we should just remove the state check here.

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -3207,8 +3207,8 @@ dhcp_handledhcp(struct interface *ifp, struct bootp *bootp, size_t bootp_len,
 
 	if (has_option_mask(ifo->requestmask, DHO_IPV6_PREFERRED_ONLY)) {
 		if (get_option_uint32(ifp->ctx, &v6only_time, bootp, bootp_len,
-		    DHO_IPV6_PREFERRED_ONLY) == 0 &&
-		    (state->state == DHS_DISCOVER || state->state == DHS_REBOOT))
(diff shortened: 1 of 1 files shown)


rsmarples:
https://datatracker.ietf.org/doc/html/rfc8925#section-3.2

```
If the client is in any other state, it SHOULD continue to use the assigned IPv4 address until further DHCPv4 reconfiguration events.
```

So we need to add the NONE state to the list, not just remove the list.

rsmarples (changes requested):
https://datatracker.ietf.org/doc/html/rfc8925#section-3.2

```
If the client is in any other state, it SHOULD continue to use the assigned IPv4 address until further DHCPv4 reconfiguration events.
```

So we need to add the NONE state to the list, not just remove the list.

taoyl-g:
Thanks for the RFC pointer. I've updated the patch.

rsmarples (approved):

rsmarples:
Thanks!

Issue #318, opened 2024-04-26
time_offset v4 option is wrong signedness 


ColinMcInnes:
dhcpcd/src/dhcpcd-definitions.conf

time_offset (dhcp option 2) is incorrectly trying to encode/parse as uint, it
needs to be int so we can do western timezone offsets (negative).

```
##############################################################################
# DHCP RFC2132 options unless otheriwse stated
define 1	request ipaddress	subnet_mask
# RFC3442 states that the CSR has to come before all other routes
# For completeness we also specify static routes then routers
define 121	rfc3442			classless_static_routes
define 2	uint32			time_offset                          <<<< should be int as per RFC-2132
define 3	request array ipaddress	routers
define 4	array ipaddress		time_servers 
define 5	array ipaddress		ien116_name_servers
define 6	array ipaddress		domain_name_servers

```

ColinMcInnes:
Resolved in Pull Request #319

rsmarples:
Thanks for the patch!

PR #319, opened 2024-04-26
Fix time_offset to be int to match RFC-2132

ColinMcInnes:
https://datatracker.ietf.org/doc/html/rfc2132#section-3.4

time_offset (dhcp v4 option 2) should be int32, not uint32, as it can indicate a time offset east (positive) or west (negative) of zero meridian.

Fixes issue #318


rsmarples:
Good spot, thanks!

PR #340, opened 2024-07-12
IPv4LL: Restart ARP probling on address conflict

LeoRuan:
When IPv4LL address conflict is detected, it is failed to restart IPv4LL
since IPv4LL is running. The commit fixes the problem by restarting ARP
probing instead of restarting IPv4LL.

### DIFF
--- a/src/ipv4ll.c
+++ b/src/ipv4ll.c
@@ -50,6 +50,12 @@
 #include "sa.h"
 #include "script.h"
 
+#ifndef KERNEL_RFC5227
+static void ipv4ll_start_arp(void *arg);
+#else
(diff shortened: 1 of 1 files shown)


rsmarples:
This looks ok so far. Sorry for the delay, I have an issue with my ARP testing tool for dhcpcd which is taking time to resolve.
Once done and I've verified one troublesome OS I'll merge this in.

rsmarples:
@LeoRuan I've added a commit to make it simpler and fix modern NetBSD kernels that support RFC 5227. Can you test to make sure things still work for you?

LeoRuan:
@rsmarples The change works well on my platform. Thanks!

PR #342, opened 2024-07-23
DHCP: Handle option 108 correctly when receiving 0.0.0.0 OFFER

taoyl-g:
According to RFC8925 section 3.3.1, when the server supports both option 108 (IPv6-Only Preferred) and option 116 (Auto-Configure), and the client only sends IPv6-Only Preferred option, then the server SHOULD return 0.0.0.0 as the offered address, and not setting the Auto-Configure option.

However, in our current client code, the IPv6-Only Preferred option in a 0.0.0.0 OFFER is only handled correctly when the Auto-Configure option is present. This patch fixes this issue.

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -774,7 +774,7 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 	}
 
 	if (ifo->options & DHCPCD_BROADCAST &&
-	    bootp->ciaddr == 0 &&
+	    bootp->ciaddr == INADDR_ANY &&
 	    type != DHCP_DECLINE &&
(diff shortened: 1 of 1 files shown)


taoyl-g:
Looks good, Thanks!

Btw, it looks like `ipv4llonly` is not really used out of the switch statement so we don't need it and can keep the old code of using `tmp`?

rsmarples:
> Looks good, Thanks!
> 
> Btw, it looks like `ipv4llonly` is not really used out of the switch statement so we don't need it and can keep the old code of using `tmp`?

I removed the ipv4llonly var. I  was planing on using it, but changed my mind.
Thanks for the patch, merged.

PR #346, opened 2024-07-29
DHCP: No longer set interface mtu

### LINKED ISSUE #345: MTU_MIN, MTU_MAX
### Discussed in https://github.com/NetworkConfiguration/dhcpcd/discussions/344

<div type='discussions-op-text'>

<sup>Originally posted by **jvfranklin** July 27, 2024</sup>
In src/dhcp.c make_message(), there's this code enforcing a minimum MTU:

else if (mtu < MTU_MIN) {
		if (if_setmtu(ifp, MTU_MIN) == -1)

MTU_MIN is defined in src/dhcp-common.h:
#define	IP_UDP_SIZE		  28
#define	MTU_MAX			1500 - IP_UDP_SIZE
#define	MTU_MIN			 576 + IP_UDP_SIZE

I work on an embedded Linux system that allows a minimum MTU on the link of 576. It appears from my initial testing that dhcpcd is able to send and receive DHCP messages without any issue if I change MTU_MIN=576. Our previous DHCP client was also able to handle this, as well. I don't see any fragmentation in the packet captures. The ethernet frames are less than 400 bytes, at least for my environment.

Is there a history on this code that requires MTU_MIN to be 604?

Also wondering about the MTU_MAX, though I haven't explored its effect in our product yet.</div>

rsmarples:
We've been enforcing an interface MTU that is slightly larger than the minimum for some time.
Instead, log an error than the MTU is smaller than the minimum to send a BOOTP message.

The DHCP MTU is only used when adding routes as setting the interface MTU can cause a PHY reset which is bad.

Fixes #345

### DIFF
--- a/src/dhcp-common.h
+++ b/src/dhcp-common.h
@@ -46,11 +46,6 @@
 #define NS_MAXLABEL MAXLABEL
 #endif
 
-/* Max MTU - defines dhcp option length */
-#define	IP_UDP_SIZE		  28
-#define	MTU_MAX			1500 - IP_UDP_SIZE
-#define	MTU_MIN			 576 + IP_UDP_SIZE
-
 #define	OT_REQUEST		(1 << 0)
 #define	OT_UINT8		(1 << 1)
 #define	OT_INT8			(1 << 2)
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -97,6 +97,8 @@
 __CTASSERT(sizeof(struct ip)		== 20);
 __CTASSERT(sizeof(struct udphdr)	== 8);
 __CTASSERT(sizeof(struct bootp)		== 300);
+#define IP_UDP_SIZE	sizeof(struct ip) + sizeof(struct udphdr)
+#define BOOTP_MIN_MTU	IP_UDP_SIZE + sizeof(struct bootp)
 
 struct dhcp_op {
 	uint8_t value;
@@ -676,6 +678,8 @@ dhcp_get_mtu(const struct interface *ifp)
 	    get_option_uint16(ifp->ctx, &mtu,
 			      state->new, state->new_len, DHO_MTU) == -1)
 		return 0;
+	if (mtu < IPV4_MMTU)
+		return IPV4_MMTU;
 	return mtu;
 }
 
@@ -740,19 +744,24 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 	uint8_t *auth, auth_len;
 #endif
 
-	if ((mtu = if_getmtu(ifp)) == -1)
+	/* We could take the DHCPv6 approach and work out the
+	 * message length up front rather than this big hammer approach. */
+	if ((mtu = if_getmtu(ifp)) == -1) {
 		logerr("%s: if_getmtu", ifp->name);
-	else if (mtu < MTU_MIN) {
-		if (if_setmtu(ifp, MTU_MIN) == -1)
-			logerr("%s: if_setmtu", ifp->name);
-		mtu = MTU_MIN;
+		return -1;
+	}
+	if ((size_t)mtu < BOOTP_MIN_MTU) {
+		logerr("%s: interface mtu is too small (%d<%zu)",
+		    ifp->name, mtu, BOOTP_MIN_MTU);
+		return -1;
 	}
 
-	if (ifo->options & DHCPCD_BOOTP)
-		bootp = calloc(1, sizeof (*bootp));
-	else
+	if (ifo->options & DHCPCD_BOOTP) {
+		bootp = calloc(1, sizeof(*bootp));
+	} else {
 		/* Make the maximal message we could send */
-		bootp = calloc(1, (size_t)(mtu - IP_UDP_SIZE));
+		bootp = calloc(1, (size_t)mtu - IP_UDP_SIZE);
+	}
 
 	if (bootp == NULL)
 		return -1;
@@ -797,7 +806,7 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 		return sizeof(*bootp);
 
 	p = bootp->vend;
-	e = (uint8_t *)bootp + (mtu - IP_UDP_SIZE) - 1; /* -1 for DHO_END */
+	e = (uint8_t *)bootp + ((size_t)mtu - IP_UDP_SIZE - 1/* DHO_END */);
 
 	ul = htonl(MAGIC_COOKIE);
 	memcpy(p, &ul, sizeof(ul));
@@ -924,7 +933,7 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 			AREA_CHECK(2);
 			*p++ = DHO_MAXMESSAGESIZE;
 			*p++ = 2;
-			sz = htons((uint16_t)(mtu - IP_UDP_SIZE));
+			sz = htons((uint16_t)((size_t)mtu - IP_UDP_SIZE));
 			memcpy(p, &sz, 2);
 			p += 2;
 		}
--- a/src/if-options.c
+++ b/src/if-options.c
@@ -1218,7 +1218,7 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 			if (p == NULL)
 				break;
 			ifo->mtu = (unsigned int)strtou(p, NULL, 0,
-			    MTU_MIN, MTU_MAX, &e);
+			    IPV4_MMTU, UINT_MAX, &e);
 			if (e) {
 				logerrx("invalid MTU %s", p);
 				return -1;
--- a/src/if.c
+++ b/src/if.c
@@ -852,27 +852,18 @@ if_loopback(struct dhcpcd_ctx *ctx)
 }
 
 int
-if_domtu(const struct interface *ifp, short int mtu)
+if_getmtu(const struct interface *ifp)
 {
-	int r;
-	struct ifreq ifr;
-
 #ifdef __sun
-	if (mtu == 0)
-		return if_mtu_os(ifp);
-#endif
+	return if_mtu_os(ifp);
+#else
+	struct ifreq ifr = { .ifr_mtu = 0 };
 
-	memset(&ifr, 0, sizeof(ifr));
 	strlcpy(ifr.ifr_name, ifp->name, sizeof(ifr.ifr_name));
-	ifr.ifr_mtu = mtu;
-	if (mtu != 0)
-		r = if_ioctl(ifp->ctx, SIOCSIFMTU, &ifr, sizeof(ifr));
-	else
-		r = pioctl(ifp->ctx, SIOCGIFMTU, &ifr, sizeof(ifr));
-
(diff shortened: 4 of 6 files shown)


mtremer:
Hello Roy,

thank you for your work on this. At IPFire, we are happy long-term users of dhcpcd, but I am not sure if we need to make some changes due to this patch.

> The DHCP MTU is only used when adding routes as setting the interface MTU can cause a PHY reset which is bad.

I agree that resetting the PHY is a problem. We have seen this here and there over time, but it has not been a major problem for our user base in a while.

In theory we can live with setting the MTU when creating the routes, but I wanted to check with you whether there are some follow-up problems that might need addressing:

* Path MTU discovery and clamp MSS to Path MTU are two things that we are using here or there. I always assumed that the kernel will use the interface's MTU and make according decisions. Since it now has to look up the route (which it would have to do anyways to find the egress interface) this will work fine (https://git.ipfire.org/?p=thirdparty/kernel/linux.git;a=blob;f=net/netfilter/xt_TCPMSS.c;h=116a885adb3cd6e93f82ee13f988e2dc0bee29f1;hb=c763c43396883456ef57e5e78b64d3c259c4babc#l62). In user space (i.e. OpenVPN) I don't know. We are not a big user of this  as Path MTU discovery doesn't work well in practise, but I wanted to flag that this might potentially become a problem.
* Additionally we allow users to configure static routes. Those might point at a gateway on the network that is configured with DHCP. Since we don't know the MTU, we can only configure them in the classic way without the MTU and we might send out packets where this route matches and therefore the requested MTU is not applied. Setting the MTU on there interface guaranteed before that we will never send any packets larger than the MTU no matter what.
* Last but not least, we use QoS in form of CAKE and fq_codel. Both set up themselves using the device MTU which will now be different from what is actually being used. The crucial function is psched_mtu (https://git.ipfire.org/?p=thirdparty/kernel/linux.git;a=blob;f=include/net/pkt_sched.h;hb=c763c43396883456ef57e5e78b64d3c259c4babc#l137) which for fq_codel will set the quantum parameter and for CAKE will be used to compute the packet overhead due to fragmentation. Miscalculating those values will make QoS less efficient the more the route MTU differs from the interface MTU.
* Maybe there are more things. I could think of IPsec which also encapsulates packets and therefore wants to know the MTU, but since that is living in the kernel, this should be fine using the route option. I did not check this.

I would like to see if we can either revert this change, make it configurable or at least export the value to the shell script so that we can potentially adjust anything manually. I agree that the PHY reset is undesirable, but is that not rather a driver problem that needs to be fixed instead of implementing this workaround which has these disadvantages?

I am happy to work on the implementation but before I start, I would like to hear your thoughts on this.

Up to the latest release, we had this patch in place (https://git.ipfire.org/?p=ipfire-2.x.git;a=blob;f=src/patches/dhcpcd-10.0.2-Allow-free-selection-of-MTU-by-the-user.patch;h=69a35daf5036ad388204f010dcb950d49e4bcc93;hb=4c672e3b9692927d4d3319cb25283098b9075a46).

Best,
-Michael

rsmarples:
Hi Michael

> thank you for your work on this. At IPFire, we are happy long-term users of dhcpcd, but I am not sure if we need to make some changes due to this patch.

Glad you're long term uses, always happy to make changes.

> > The DHCP MTU is only used when adding routes as setting the interface MTU can cause a PHY reset which is bad.
> 
> I agree that resetting the PHY is a problem. We have seen this here and there over time, but it has not been a major problem for our user base in a while.
> 
> In theory we can live with setting the MTU when creating the routes, but I wanted to check with you whether there are some follow-up problems that might need addressing:

OK, before we talk about these points, dhcpcd has not set the interface MTU for almost 10 years now.
https://github.com/NetworkConfiguration/dhcpcd/commit/ca6cdf5847cda720b65793ea6827b1b373c62382

> 
>     * Path MTU discovery and clamp MSS to Path MTU are two things that we are using here or there. I always assumed that the kernel will use the interface's MTU and make according decisions. Since it now has to look up the route (which it would have to do anyways to find the egress interface) this will work fine (https://git.ipfire.org/?p=thirdparty/kernel/linux.git;a=blob;f=net/netfilter/xt_TCPMSS.c;h=116a885adb3cd6e93f82ee13f988e2dc0bee29f1;hb=c763c43396883456ef57e5e78b64d3c259c4babc#l62). In user space (i.e. OpenVPN) I don't know. We are not a big user of this  as Path MTU discovery doesn't work well in practise, but I wanted to flag that this might potentially become a problem.

I can't speak for Linux, but in BSD when we need to look at any MTU we always use the lowest of the outbound route or the interface MTU. We always have a route and an interface, so there is no extra lookup needed.

> 
>     * Additionally we allow users to configure static routes. Those might point at a gateway on the network that is configured with DHCP. Since we don't know the MTU, we can only configure them in the classic way without the MTU and we might send out packets where this route matches and therefore the requested MTU is not applied. Setting the MTU on there interface guaranteed before that we will never send any packets larger than the MTU no matter what.

The user can set a static mtu as well, which should be applied to the routes dhcpcd configures.

>     * Last but not least, we use QoS in form of CAKE and fq_codel. Both set up themselves using the device MTU which will now be different from what is actually being used. The crucial function is psched_mtu (https://git.ipfire.org/?p=thirdparty/kernel/linux.git;a=blob;f=include/net/pkt_sched.h;hb=c763c43396883456ef57e5e78b64d3c259c4babc#l137) which for fq_codel will set the quantum parameter and for CAKE will be used to compute the packet overhead due to fragmentation. Miscalculating those values will make QoS less efficient the more the route MTU differs from the interface MTU.

That sounds like a CAKE issue?
Anyway, the important thing is that DHCP MTU now behaves exactly like IPv6 in-kernel RA MTU which only sets MTU on routes.
So if it is a CAKE issue then it needs to be resolved outside of dhcpcd for kernel IPv6 RA MTU anyway.

>     * Maybe there are more things. I could think of IPsec which also encapsulates packets and therefore wants to know the MTU, but since that is living in the kernel, this should be fine using the route option. I did not check this.

See above for CAKE.

> I would like to see if we can either revert this change, make it configurable or at least export the value to the shell script so that we can potentially adjust anything manually. I agree that the PHY reset is undesirable, but is that not rather a driver problem that needs to be fixed instead of implementing this workaround which has these disadvantages?
> 
> I am happy to work on the implementation but before I start, I would like to hear your thoughts on this.

This change just removed changing the MTU to the "minimum" size needed for DHCP.
It never changed the interface MTU for the MTU from a DHCP message. That was removed over 10 years ago as I said earlier.

> Up to the latest release, we had this patch in place (https://git.ipfire.org/?p=ipfire-2.x.git;a=blob;f=src/patches/dhcpcd-10.0.2-Allow-free-selection-of-MTU-by-the-user.patch;h=69a35daf5036ad388204f010dcb950d49e4bcc93;hb=4c672e3b9692927d4d3319cb25283098b9075a46).

I don't think your patch is needed anymore as we now accept UINT_MAX MTU which is bigger than what you changed to.

mtremer:
> Glad you're long term uses, always happy to make changes.

Currently we would not require anything as dhcpcd has been working like a charm :)

> OK, before we talk about these points, dhcpcd has not set the interface MTU for almost 10 years now. [ca6cdf5](https://github.com/NetworkConfiguration/dhcpcd/commit/ca6cdf5847cda720b65793ea6827b1b373c62382)

Oh, this is a lot older than I remembered. I do recall that this change gave us some trouble and I thought that we still had some sort of workaround that we would employ to change the interface MTU.

Granted, I don't think that there are many networks that run on anything else but 1500. There might be some fibre modems that encapsulate packets later on and might want 1492 or even smaller, but I don't believe these things are very common. Even then, connections should just work even with 1500 if a router further down the line performs any fragmentation.

I must have misread the commit then. Regardless, we have discovered some very minor problem here...

> > * Additionally we allow users to configure static routes. Those might point at a gateway on the network that is configured with DHCP. Since we don't know the MTU, we can only configure them in the classic way without the MTU and we might send out packets where this route matches and therefore the requested MTU is not applied. Setting the MTU on there interface guaranteed before that we will never send any packets larger than the MTU no matter what.
> > ```
> 
> The user can set a static mtu as well, which should be applied to the routes dhcpcd configures.

Yes, we have an option for that and I do remember testing this with the changes from ~10 years ago. The static routes are just set by a different script in a different route table. This can all be changed since in this case we know the MTU from the user. My problem is that I don't seem to have a way to find the MTU when it has been changed by the DHCP server. The variable interface_mtu= is empty in the script (or I am too dumb to test things right or I have a cable modem that does not specify a MTU).

> > * Last but not least, we use QoS in form of CAKE and fq_codel. Both set up themselves using the device MTU which will now be different from what is actually being used. The crucial function is psched_mtu (https://git.ipfire.org/?p=thirdparty/kernel/linux.git;a=blob;f=include/net/pkt_sched.h;hb=c763c43396883456ef57e5e78b64d3c259c4babc#l137) which for fq_codel will set the quantum parameter and for CAKE will be used to compute the packet overhead due to fragmentation. Miscalculating those values will make QoS less efficient the more the route MTU differs from the interface MTU.
> > ```
> 
> That sounds like a CAKE issue? Anyway, the important thing is that DHCP MTU now behaves exactly like IPv6 in-kernel RA MTU which only sets MTU on routes. So if it is a CAKE issue then it needs to be resolved outside of dhcpcd for kernel IPv6 RA MTU anyway.

This is a very strong point that I did not even consider. Thank you for this, I will reach out to the CAKE people and ask if they think that this might cause a problem.

> > I would like to see if we can either revert this change, make it configurable or at least export the value to the shell script so that we can potentially adjust anything manually. I agree that the PHY reset is undesirable, but is that not rather a driver problem that needs to be fixed instead of implementing this workaround which has these disadvantages?
> > I am happy to work on the implementation but before I start, I would like to hear your thoughts on this.
> 
> This change just removed changing the MTU to the "minimum" size needed for DHCP. It never changed the interface MTU for the MTU from a DHCP message. That was removed over 10 years ago as I said earlier.

We needed this because there was only a small corridor for the MTU and I remember that this patch was needed to virtually allow the user to pick whatever they want.

So, thank you for your help for now. I will reach out to the CAKE people and see if this is actually a problem here or if there is a way to get around this.

rsmarples:
> Yes, we have an option for that and I do remember testing this with the changes from ~10 years ago. The static routes are just set by a different script in a different route table. This can all be changed since in this case we know the MTU from the user. My problem is that I don't seem to have a way to find the MTU when it has been changed by the DHCP server. The variable interface_mtu= is empty in the script (or I am too dumb to test things right or I have a cable modem that does not specify a MTU).

You should see old and new values for mtu when changed by DHCP.
I see this in my testing.

```
reason=BOUND
interface=vtnet0
interface_mtu=1492
```

> > This change just removed changing the MTU to the "minimum" size needed for DHCP. It never changed the interface MTU for the MTU from a DHCP message. That was removed over 10 years ago as I said earlier.
> 
> We needed this because there was only a small corridor for the MTU and I remember that this patch was needed to virtually allow the user to pick whatever they want.

And that corridor has been removed, the user is free to pick whatever.

> So, thank you for your help for now. I will reach out to the CAKE people and see if this is actually a problem here or if there is a way to get around this.

Welcome.
But open a new issue if needed.
You can reference this if needed.

PR #357, opened 2024-08-28
dhcp: get_option_uint32/16 only accept options with correct len

taoyl-g:
RFC8925 mentions "The client MUST ignore the IPv6-Only Preferred option if the length field value is not 4."

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -304,7 +304,7 @@ get_option_uint32(struct dhcpcd_ctx *ctx,
 	uint32_t d;
 
 	p = get_option(ctx, bootp, bootp_len, option, &len);
-	if (!p || len < (ssize_t)sizeof(d))
+	if (!p || len != (ssize_t)sizeof(d))
 		return -1;
 	memcpy(&d, p, sizeof(d));
 	if (i)
@@ -321,7 +321,7 @@ get_option_uint16(struct dhcpcd_ctx *ctx,
 	uint16_t d;
 
 	p = get_option(ctx, bootp, bootp_len, option, &len);
-	if (!p || len < (ssize_t)sizeof(d))
+	if (!p || len != (ssize_t)sizeof(d))
 		return -1;
 	memcpy(&d, p, sizeof(d));
 	if (i)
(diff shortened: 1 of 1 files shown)


taoyl-g:
I'm not modifying get_option_uint8 now as it's also used to read variant length options. Please suggest if you have other preferred way to fix this (such as adding another helper that does not verify the length)

rsmarples:
> I'm not modifying get_option_uint8 now as it's also used to read variant length options. Please suggest if you have other preferred way to fix this (such as adding another helper that does not verify the length)

That would be the correct way I think.
We have two calls that use the uint8 function just to check the existence of an option so we could have a new function that does just that.

taoyl-g:
Enforcing len==1 in get_option_uint8 as well, and using get_option instead in the mentioned occurrences.

src/dhcp.c
```diff
@@ -3158,8 +3158,7 @@ dhcp_handledhcp(struct interface *ifp, struct bootp *bootp, size_t bootp_len,
 	/* Ensure that no reject options are present */
 	for (i = 1; i < 255; i++) {
 		if (has_option_mask(ifo->rejectmask, i) &&
-		    get_option_uint8(ifp->ctx, &tmp,
-		    bootp, bootp_len, (uint8_t)i) == 0)
+		    get_option(ifp->ctx, bootp, bootp_len, (uint8_t)i, NULL))
```
rsmarples, line 3161: This should be `!get_option` as we are testing the option to not exist.

src/dhcp.c
```diff
@@ -3207,8 +3206,7 @@ dhcp_handledhcp(struct interface *ifp, struct bootp *bootp, size_t bootp_len,
 	/* Ensure that all required options are present */
 	for (i = 1; i < 255; i++) {
 		if (has_option_mask(ifo->requiremask, i) &&
-		    get_option_uint8(ifp->ctx, &tmp,
-		    bootp, bootp_len, (uint8_t)i) != 0)
+		    !get_option(ifp->ctx, bootp, bootp_len, (uint8_t)i, NULL))
```
rsmarples, line 3209: This should be `get_option` as we are testing the option to exist.

src/dhcp.c
taoyl-g: We are testing the option to exist here, isn't it? (It's `LOG_WARNING` and `return` in the `if` body)

src/dhcp.c
taoyl-g: Similar with above, I think we are testing the option to not exist here. (Note that `get_option_uint8` returns 0 on existence and -1 on non-existence, while `get_option` returns NULL on non-existence and non-zero on existence)

rsmarples:
Heh, I should look at the larger picture some more. Thanks for your work and replies.

PR #364, opened 2024-09-12
dhcpcd: stdout output sometimes empty when redirected to a file

diego-santacruz:
Running dhpcd --dumplease with input from stdin and redirecting stdout to a file outputs nothing to the output. The reason is that printf / fprintf is used to write the output but it is not explicitly flushed when done, or before exiting, and stdout to files is fully buffered by default while stdout to terminals is line buffered by default.

This adds calls to fflush() at the end of dhcp_print_option_encoding() and script_dump(), plus one just before exiting dhcpcd.

### DIFF
--- a/src/dhcp-common.c
+++ b/src/dhcp-common.c
@@ -136,6 +136,7 @@ dhcp_print_option_encoding(const struct dhcp_opt *opt, int cols)
 	if (opt->type & OT_NOREQ)
 		printf(" norequest");
 	putchar('\n');
+	fflush(stdout);
 }
 
 struct dhcp_opt *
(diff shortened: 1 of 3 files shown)


cross:
Curious, why does the output not flush when dhcpcd exits?  exit() should close stdout, which will implicitly flush it, right?

diego-santacruz:
> Curious, why does the output not flush when dhcpcd exits? exit() should close stdout, which will implicitly flush it, right?

Good question, I think it is because with `--dumplease` the processing in `main()` pretends that the process forked and by the end of main `_exit()` is called if the process forked, for which flushing of stdio streams is platform-dependent and I guess not being done in Linux with glibc. The code is a bit difficult to follow though, so I may be mistaken.

cross:
Ahh.  It uses _exit().  Then that explains it.  _Why_ it uses _exit() I don't know, but that answers my question.  Thanks.

PR #374, opened 2024-10-04
DHCPv6: Add support for sending Option 17 (VSIO)

spoljak-ent:
Hi,
I made a patch for the support of sending DHCPv6 Option 17.
The options can be added similarly to the current vendor option 43.

Just add in config file as many options as needed like this:
vsio ent_num opt_num,option

Please comment and review.
Thanks.

### DIFF
--- a/src/ipv6.c
+++ b/src/ipv6.c
@@ -2331,6 +2331,7 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 			sa_in6_init(&rt->rt_dest, &rinfo->prefix);
 			sa_in6_init(&rt->rt_netmask, &netmask);
 			sa_in6_init(&rt->rt_gateway, &rap->from);
+			rt->rt_dflags |= RTDF_RA;
 #ifdef HAVE_ROUTE_PREF
 			rt->rt_pref = ipv6nd_rtpref(rinfo->flags);
 #endif


src/dhcp6.c
```diff
@@ -336,6 +336,66 @@ dhcp6_makevendor(void *data, const struct interface *ifp)
...
+	uint8_t *p;
+	const struct vsio6 *vsio6;
+	struct dhcp6_option o;
+
+	ifo = ifp->options;
+	len = sizeof(uint32_t); /* IANA PEN */
+	vlen = 0;
+
+	for (i = 0, vsio6 = ifo->vsio6; i < ifo->vsio6_len; 
+		i++, vsio6++) {
```
rsmarples, line 354: Rather than increment two things, we can just increment vsio6.
For this to work we need to work out the address of the final pointer using vsio6_len.
`*vsio6_endp = ifo->vsio6 + ifo->vsio6_len` and then compare.
I would front load this in the variable declaration at the top to reduce the number of  lines the for loop takes up.

src/dhcp6.c
```diff
@@ -336,6 +336,66 @@ dhcp6_makevendor(void *data, const struct interface *ifp)
...
+			if (vsio6->en != en)
+				continue;
+			vlen += 2 * sizeof(uint16_t) + vsio6->len;
+		}
```
rsmarples, line 358: This should be one indent to the left.

src/dhcp6.c
```diff
@@ -336,6 +336,66 @@ dhcp6_makevendor(void *data, const struct interface *ifp)
...
+		o.code = htons(D6_OPTION_VENDOR_OPTS);
+		o.len = htons((uint16_t)len);
+		memcpy(p, &o, sizeof(o));
+		p += sizeof(o);
+		pen = htonl(en);
+		memcpy(p, &pen, sizeof(pen));
+		p += sizeof(pen);
+
+		for (i = 0, vsio6 = ifo->vsio6; i < ifo->vsio6_len;
+			i++, vsio6++) {
```
rsmarples, line 382: Same as above.

rsmarples (changes requested):
Can we wrap this in `#ifndef SMALL` please so we don't bloat the code for installations where size matters?

spoljak-ent:
> Can we wrap this in `#ifndef SMALL` please so we don't bloat the code for installations where size matters?

No problem... Just to be sure, the whole implementation or something specific?

rsmarples:
> > Can we wrap this in `#ifndef SMALL` please so we don't bloat the code for installations where size matters?
> 
> No problem... Just to be sure, the whole implementation or some specific lines?

Whole implementation please.

spoljak-ent:
> > > Can we wrap this in `#ifndef SMALL` please so we don't bloat the code for installations where size matters?
> > 
> > 
> > No problem... Just to be sure, the whole implementation or some specific lines?
> 
> Whole implementation please.
Sure thing.
Had to pull recent changes and everything was overwritten... I've created a new pull request for this:
https://github.com/NetworkConfiguration/dhcpcd/pull/382

PR #377, opened 2024-10-07
DHCPv4: Add support for sending Option 125 (VIVSO)

spoljak-ent:
Hi,

Similar to the DHCPv6 Option 17 patch, I made one for DHCPv4 Option 125.
The syntax for the config file is identical with the first argument starting with vivso.
So, it's:

vivso ent_num opt_num,opt

As with the previous one, please comment and review.
Thanks :)

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -723,6 +723,66 @@ dhcp_message_add_addr(struct bootp *bootp,
 	return 0;
 }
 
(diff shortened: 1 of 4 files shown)


ColinMcInnes:
I recommend adding an example to src/dhcpcd-definitions.conf

spoljak-ent:
Implemented with
#383

Issue #74, opened 2021-11-18
Allow multiple subopt with same code


wrjiafang:
Sometimes we will send the message with different dhcp option but with same code, so we try to add this function to allow multiple sub-opt. Please help to review this patch and give your suggestions, thanks very much !
Here is the patch:
[0001-Allow-multiple-subopt-with-same-code.patch.txt](https://github.com/NetworkConfiguration/dhcpcd/files/7560775/0001-Allow-multiple-subopt-with-same-code.patch.txt)

rsmarples:
Could you post an example message containing this please?

FangJoanna:
Thanks very much for your quick response, I will attach this kind of message, please wait for a while.

wrjiafang:
Thanks very much for your quick reply. I will attach this kind of message, please wait for a while.

wrjiafang:
Sorry for a late response.

Please check the attached file:
![image](https://user-images.githubusercontent.com/12740056/144581075-3bc42e4d-da5e-47ed-bf95-64aa5658b61b.png)
Our dhcp server will send the message as attached.
The server will sends multiple instances of Verdor sub-options within Option 17/43/125, the attached is an example of Option 125. Right now, dhcp client only handle one instances.

rsmarples:
Thanks for the example. Could you submit a PR for review please?
From the looks of the patch, it's concatenating the options together so the end result is `option125="ac114102ac114101"` yes?

perkelix:
Does the above still apply to 10.5.2?

PR #378, opened 2024-10-07
DHCP: Handling multiple vendor sub-options with the same code number

spoljak-ent:
This is the same patch which was posted here:
https://github.com/NetworkConfiguration/dhcpcd/issues/74

### DIFF
--- a/src/dhcp-common.c
+++ b/src/dhcp-common.c
@@ -874,18 +874,20 @@ dhcp_envoption(struct dhcpcd_ctx *ctx, FILE *fp, const char *prefix,
     const uint8_t *, size_t, struct dhcp_opt **),
     const uint8_t *od, size_t ol)
 {
-	size_t i, eos, eol;
+	size_t i, eos, eol, buflen, tol;
 	ssize_t eo;
 	unsigned int eoc;
-	const uint8_t *eod;
+	const uint8_t *eod, *tod;
 	int ov;
 	struct dhcp_opt *eopt, *oopt;
 	char *pfx;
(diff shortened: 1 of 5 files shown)


rsmarples:
@spoljak-ent we no longer need this do we?

rsmarples:
Ah, it's mixed in with the others. Apologies, I had not actually realised you had already done a PR for the DHCP 125 option.

But to answer the question, for DHCPv6 they should be indexed
```
vsio_en=155
vsio_str='foo'
vsio_str1='bar'
vsio1_en=155
vsio1_str='hello'
vsio1_str1='there'
```

For DHCPv4 they *should* be concatenated in in accordance with RFC3396
```
vsio_en=155
vsio_str='foobar'
vsio1_en=155
vsio1_str='hellothere'
```

But equally that could be driven by the EN - we just don't know.
To make it simple, I would just have both work like DHCPv6 (which I think they should do now, but never really tested) and maybe
add an option in the future to concatenate DHCP.

Does that help?

spoljak-ent:
It does help, thanks for the info.
Don't know if there were any changes regarding this behavior in the meantime (since this issue (#74) is open for a long time)?
The issue was that multiple sub-options (for option 125 at least) with the same code would only get one of them printed out to env. This patch (from the aforementioned open issue) concatenates them so they all get printed.

spoljak-ent:
Just tested this again on the latest version with the server sending Option 125 with two different strings with the same option number and it only prints one of the strings.... it doesn't concatenate them.
This patch fixes the issue.

Issue #343, opened 2024-07-25
IPv6 RA should respect and utilize kernel stable-privacy and tempaddr flags


perkelix:
In its current form (10.0.6), dhcpcd ignores kernel settings and creates pseudo-private IPv6 addresses. If `net.ipv6.conf.*.addr_gen_mode=3` or `net.ipv6.conf.*.use_tempaddr=2` (Linux) are set by procps during bootup, dhcpcd gnores them.

This is undesirable for a number of reasons.

What would be a better implementation is to first try enabling stable-privacy and, if that fails, enable temporary addresses. The later has been implemented ages ago and will thus probably always succeed, and it would work well as a failsafe for the former, whose implementation is more recent on most kernels.

This implementation should also respect both options if they have already been configured during bootup.

PR #380, opened 2024-10-09
Don't configure an IPv6 link-local address if noipv6rs is set

jvfranklin:
The configuration I'm using uses the noipv6rs option to allow the Linux kernel to do router discovery, slaac, etc. I'm seeing dhcpcd configuring an EUI64 format link-local address unnecessarily. This change stops dhcpcd from doing that.

### DIFF
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -2173,6 +2173,11 @@ if_setup_inet6(const struct interface *ifp)
 	int ra;
 	char path[256];
 
+	/* If not doing autoconf, don't disable the kernel from doing it.
+	 * If we need to, we should have another option actively disable it. */
+	if (!(ifp->options->options & DHCPCD_IPV6RS))
+		return;
+
 	/* Modern linux kernels can make a stable private address.
 	 * However, a lot of distros ship newer kernel headers than
 	 * the kernel itself so we sweep that error under the table
@@ -2180,11 +2185,6 @@ if_setup_inet6(const struct interface *ifp)
 	if (if_disable_autolinklocal(ctx, ifp->index) == -1 &&
 	    errno != ENODEV && errno != ENOTSUP && errno != EINVAL)
 		logdebug("%s: if_disable_autolinklocal", ifp->name);
-
-	/* If not doing autoconf, don't disable the kernel from doing it.
-	 * If we need to, we should have another option actively disable it. */
-	if (!(ifp->options->options & DHCPCD_IPV6RS))
-		return;
 
 	snprintf(path, sizeof(path), "%s/%s/autoconf", p_conf, ifp->name);
 	ra = check_proc_int(ctx, path);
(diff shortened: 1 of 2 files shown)


jvfranklin:
This is a demo of what I'm seeing. Our product is a Linux-based system with a custom network manager daemon, so I'm just using dhcpcd as a DHCP/DHCPv6 client. If I want to use stable-privacy addresses, managed by the kernel, dhcpcd interferes with that when it starts up:

/home/root# cat /tmp/dhcpcd.conf
allowinterfaces nic1
interface nic1
noup
noipv4
ipv6ra_noautoconf
noipv6rs
inform6
/home/root# cat /proc/sys/net/ipv6/conf/nic1/addr_gen_mode
2
/home/root# ip addr show nic1
12: nic1@eth2: <BROADCAST,MULTICAST> mtu 1500 qdisc noqueue state DOWN group default qlen 1000
    link/ether 6c:3c:8c:85:23:59 brd ff:ff:ff:ff:ff:ff
/home/root# /tmp/dhcpcd -f /tmp/dhcpcd.conf
dhcpcd-10.1.0 starting
dev: loaded udev
DUID 00:01:00:01:2e:90:6e:ee:6c:3c:8c:85:23:59
Dropped protocol specifier '.link' from 'nic1.link'. Using 'nic1' (ifindex=12).
no interfaces have a carrier
/home/root# cat /proc/sys/net/ipv6/conf/nic1/addr_gen_mode
1

So now if I bring up nic1, it comes up with an EUI64 link-local instead of stable-privacy. I've also seen the interface have both, but I think it's timing dependent.

/home/root# ip link set dev nic1 up
/home/root# ip addr show nic1
12: nic1@eth2: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
    link/ether 6c:3c:8c:85:23:59 brd ff:ff:ff:ff:ff:ff
    inet6 fe80::6e3c:8cff:fe85:2359/64 scope link
       valid_lft forever preferred_lft forever

With my patch, addr_gen_mode remains set to 2, and I don't have any conflicts between the network manager and dhcpcd.

rsmarples:
Need to think about this because you can use DHCPv6 without RS/RA and DHCPv6 still requires a LL address.

I think we need a new option, `noipv6ll`, but we're also out of options for unsigned long long.
Might need to create a new option group for inet6 and move all INET6 options into it (apart from DHCPCD_IPV6).

I won't be around much this week so I'll think it over some.

rsmarples:
#343 is related to this I guess.

perkelix:
It kinda is related to #343 since it's about not touching the interface's flags unless explicitly asked to do so.

However, what this bug wants is more akin to having a configuration option for disabling DHCPv6 and letting RS do its thing without dhcpcd's intervention.

What it instead reminds me of is something I reported a long time ago (I cannot remember the bug number): in some circumstances, a superflous LL6 address that uses EUI64 gets added in addition to a LL6 address that uses dhcpcd's pseudo-privacy.

perkelix:
Have we reached a consensus on this MR?

Issue #328, opened 2024-06-05
DHCPv4  V-I Vendor Class with multiple ENs overwrites the EN values


rmarioe:
Hi,

I see when defining the vendclass two times with different EN, the first one is overwritten and the second one is actually part of the data.
In the manual I see:

[vendclass] en data
    Add the DHCPv6 Vendor Indetifying Vendor Class with the IANA assigned Enterprise Number en with the data. This option can be set more than once to add more data, but the behaviour, as per RFC 3925 is undefined if the Enterprise Number differs.

However the RFC states:

This option contains information corresponding to one or more
   Enterprise Numbers.  Multiple instances of this option may be present
   and MUST be concatenated in accordance with [RFC 3396].  An
   Enterprise Number SHOULD only occur once among all instances of this
   option.  Behavior is undefined if an Enterprise Number occurs
   multiple times.  The information for each Enterprise Number is
   treated independently, regardless or whether it occurs in an option
   with other Enterprise Numbers or in a separate option.

Does this not mean that having multiple enterprise numbers that differ should be supported but repeating an already seen enterprise number is undefined?

rsmarples:
This isn't very well done I have to confess I probably haven't implemented this very well.
Currently you cannot specify different EN values, only the first one is used.

rmario1:
One more thing is the formatting as per https://mailarchive.ietf.org/arch/msg/dhcwg/B4fNsvUR0EHxcrKDsKCf7lBrc3s/ where each instance the vendor-class-data needs to have a length. Since the packet sent does not include it (but includes the total length) the packets are seen as malformed.
Should we track this as two issues even though one commit could fix both of them?

rsmarples:
One commit can fix two issues, sure.

rsmarples:
This will be fixed in #383 with a new option.

PR #383, opened 2024-10-10
DHCPv6: Add support for sending Option 17 (VSIO)

spoljak-ent:
My apologies, some lines were not included in #ifndef SMALL

Original pull request:
https://github.com/NetworkConfiguration/dhcpcd/pull/374


src/if-options.c
```diff
@@ -896,6 +899,99 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 		}
 		ifo->userclass[0] = (uint8_t)s;
 		break;
+	case O_VENDOPT6:
+		ARG_REQUIRED;
```
rsmarples, line 903: Wrap in SMALL please.

```
#ifdef SMALL
                        logwarnx("%s: vendor options not compiled in", ifname);
                        return -1;
#else
...
#endif
```

spoljak-ent: Done

ColinMcInnes:
I recommend adding a few config examples to src/dhcpcd-definitions.conf

spoljak-ent:
> I recommend adding a few config examples to src/dhcpcd-definitions.conf

No problem, but just to double-check, isn't dhcpcd-definitions.conf just for decoding options? This improvement is actually to enable dhcpcd to send those options.

ColinMcInnes:
Good point
```
# DHCP option definitions for dhcpcd(8)
# These are used to translate DHCP options into shell variables
# for use in dhcpcd-run-hooks(8)
# See dhcpcd.conf(5) for details
```
Maybe put it in dhcpcd.conf instead.

spoljak-ent:
> Good point
> 
> ```
> # DHCP option definitions for dhcpcd(8)
> # These are used to translate DHCP options into shell variables
> # for use in dhcpcd-run-hooks(8)
> # See dhcpcd.conf(5) for details
> ```
> 
> Maybe put it in dhcpcd.conf instead.

Done, updated as requested.

src/dhcp6.c
```diff
@@ -640,6 +640,64 @@ dhcp6_delegateaddr(struct in6_addr *addr, struct interface *ifp,
...
+// DHCPv6 Option 17 (Vendor-Specific Information Option)
+static size_t
+dhcp6_makevendoropts(void *data, const struct interface *ifp, uint32_t en)
+{
+	const struct if_options *ifo;
+	size_t len, vlen;
+	uint8_t *p;
+	struct dhcp6_option o;
+
+	ifo = ifp->options;
```
rsmarples, line 653: Please assign ifo in the declaration above so there are no blank lines in the declaration. Ensure there is a blank line after the declaration block please.

src/dhcp6.c
```diff
@@ -640,6 +640,64 @@ dhcp6_delegateaddr(struct in6_addr *addr, struct interface *ifp,
...
+	const struct vsio6 *vsio6_endp = ifo->vsio6 + ifo->vsio6_len; 
+	len = sizeof(uint32_t); /* IANA PEN */
+	vlen = 0;
+
+	for (struct vsio6 *vsio6 = ifo->vsio6; vsio6 != vsio6_endp; ++vsio6) {
+		if (vsio6->en != en)
+			continue;
+		vlen += 2 * sizeof(uint16_t) + vsio6->len;
+		}
```
rsmarples, line 662: Ensure the closing brace indention matches the for loop.

src/dhcp6.c
```diff
@@ -640,6 +640,64 @@ dhcp6_delegateaddr(struct in6_addr *addr, struct interface *ifp,
...
+				continue;
+			opt = htons((uint16_t)vsio6->opt);
+			memcpy(p, &opt, sizeof(opt));
+			p += sizeof(opt);
+			hvlen = htons((uint16_t)vsio6->len);
+			memcpy(p, &hvlen, sizeof(hvlen));
+			p += sizeof(hvlen);
+			memcpy(p, vsio6->data, vsio6->len);
+			p += vsio6->len;
+			}
```
rsmarples, line 696: Fix indentation here please.

rsmarples (commented):
Looking good so for, just a few formatting nits.
I'll see if I can find the time to actually test it soon.

src/dhcp6.c
spoljak-ent: Done

src/dhcp6.c
spoljak-ent: Done

ColinMcInnes:
Looks good to me!
```
vsio 4491 1,00:20:00:21:00:22:00:24:00:25:00:26:00:27:00:3D
vsio 4491 2,"Device-class"
vsio 4491 3,123456
```
image

rsmarples:
@spoljak-ent I'd like to add a commit to your branch which simplifies things a little and will allow code re-use for the equivalent DHCP option. Can you enable maintainer commit please?

rsmarples:
@spoljak-ent @ColinMcInnes I added a change to simplify things, which looks good from here.
Was able to test it sending many EN definitions without issue. We could re-use this for DHCP equivalent as well now.

Does this look OK to everyone?

spoljak-ent:
> @spoljak-ent @ColinMcInnes I added a change to simplify things, which looks good from here. Was able to test it sending many EN definitions without issue. We could re-use this for DHCP equivalent as well now.
> 
> Does this look OK to everyone?

Looks OK from my side :)

rsmarples:
@spoljak-ent @ColinMcInnes ok, DHCP option 125 now works! It's different from option 43 in that it splits the options via EN.

Please test this before I merge, ideally later today.

spoljak-ent:
> @spoljak-ent @ColinMcInnes ok, DHCP option 125 now works! It's different from option 43 in that it splits the options via EN.
> 
> Please test this before I merge, ideally later today.

Tested with the following configs.

DHCPv6:
`vsio6 155 01,11:22:33:44:55:66`
`vsio6 155 02,2001:0db8:3c4d:0015:0000:0000:1a2f:1a2b`
`vsio6 255 01,"hello world"`
`vsio6 255 02,55:66:77:88:99:00`
`vsio6 155 03,"another string"`
`vsio6 355 05,"/3da2k/"`

The result:
image


DHCPv4:
`vsio 155 01,192.168.8.2`
`vsio 155 02,"hello world"`
`vsio 255 01,185.53.88.15`
`vsio 255 02,15:16:17:18:19:20`
`vsio 155 03,"hello world again"`
`vsio 355 05,175.33.18.19`
The result:
image

Works like a charm :)

rsmarples:
@spoljak-ent OK, I agree. I've added some code to encode as RFC3396 - not tested that it will overflow the 255 boundary yet though. If you have the time, I'd appreciate it if you could test that.

spoljak-ent:
> @spoljak-ent OK, I agree. I've added some code to encode as RFC3396 - not tested that it will overflow the 255 boundary yet though. If you have the time, I'd appreciate it if you could test that.

@rsmarples Thanks for the update... I'll test it right now.

spoljak-ent:
So, I tried 2 configurations, first one using large strings:

image

And it malforms the packet:
image

The second one using multiple smaller strings:
image

And the result was the same as for the first case... the length was set at 159 in both cases where it broke:
image

rsmarples:
Gah, you posted pictures! I'll manually copy into my config and see if I can fix it later.

spoljak-ent:
Sorry, here's the text:

_vsio 193 1,"ShortEntry"
vsio 155 2,"A medium-length entry with enough content to add to the total byte count and push the packet closer to its limit"
vsio 255 3,"A much longer entry that includes a significant amount of configuration data to test larger data sizes in DHCP packets. This is useful for scenarios that require substantial information within a single option, going well beyond typical lengths."
vsio 200 4,"Extra data"
vsio 210 5,"Additional configuration entry to exceed 255 bytes and ensure concatenation is required"_

_vsio 193 1,"ShortEntryExample01"
vsio 155 2,"ShortEntryExample02"
vsio 255 3,"ShortEntryExample03"
vsio 200 4,"ShortEntryExample04"
vsio 210 5,"ShortEntryExample05"
vsio 193 6,"ShortEntryExample06"
vsio 155 7,"ShortEntryExample07"
vsio 255 8,"ShortEntryExample08"
vsio 200 9,"ShortEntryExample09"
vsio 210 10,"ShortEntryExample10"
vsio 193 11,"ShortEntryExample11"
vsio 155 12,"ShortEntryExample12"
vsio 255 13,"ShortEntryExample13"
vsio 200 14,"ShortEntryExample14"
vsio 210 15,"ShortEntryExample15"_

rsmarples:
@spoljak-ent ok, I've pushed a fix for that, but please note this test data:

```
vsio 193 1,"ShortEntry"
vsio 255 3,"A much longer entry that includes a significant amount of configuration data to test larger data sizes in DHCP packets. This is useful for scenarios that require substantial information within a single option, going well beyond typical lengths."
```

Wireshark still thinks it's a malformed packet as it overflows by a few bytes, but the encoding looks correct to me as per https://datatracker.ietf.org/doc/html/rfc3396#section-8

I added another change to easily test small boundaries and it still looks bad in wireshark but looks correct to me.
What do you think? If you agree we could create a Wireshark ticket?

rsmarples:
image

Basically it's just considering each option singularly and not concatenating the occurrences into a whole.

spoljak-ent:
I've been looking at this a bit and from what I understood in the RFC, the split should be done by making a completely new option and taking into account everything that option needs along with concatenation:

rsmarples:
> I've been looking at this a bit locally with Wireshark... Going back to RFC3396 section 8 which you linked, from what I understood, the split should be done by making a completely new option, taking into account everything that option needs (new total length, enterprise ID and sub-opt length) and then adding the rest of the concatenation.

From section 6:
> The encoding agent MUST NOT attempt to specify any semantic information based on how the option is split.

So the split option is just code and length - anything else is just a blob to be concatenation to the previous blob. We cannot adjust any sub-opt length.

From section 7, decoding:
>In the case that a decoding agent finds a split option, it MUST treat
   the contents of that option as a single option, and the contents MUST
   be reassembled in the order that was described above under encoding
   agent behavior.

Once re-assembled the sub option lengths are now valid.
I'm pretty sure wireshark is not doing the re-assembly for this option.

Note that at the end of the initial wireshark decoded 125 option, you can see another 125 option immediately afterwards and a length.
To prove that the lengths are correct, you can change the code on line 1174 from DHO_VIVSO to 43 and wireshark  will now claim the packet is now fine as it's making no attempt to deconstruct the 125 option.

rsmarples:
Referenced upstream here: https://gitlab.com/wireshark/wireshark/-/issues/20201

spoljak-ent:
Thanks! :)
When I find some time, I'll make an update for Vendor Class options - DHCPv4 Option 124 and DHCPv6 Option 16 using the same logic implemented just now (as those options are nearly identical to 125 and 17 respectively).
For now, those aren't working properly when using multiple enterprise IDs (as it's mentioned in https://github.com/NetworkConfiguration/dhcpcd/issues/328).

spoljak-ent:
@rsmarples 
For the implementation of Vendor Class Options 16 and 124, shall I wrap them also in #ifdef SMALL ?

src/dhcp.c
```diff
@@ -723,6 +723,76 @@ dhcp_message_add_addr(struct bootp *bootp,
...
+	while (len != 0) {
+		if (ctx->len == NULL || *ctx->len == RFC3396_BOUNDARY) {
+			if (ctx->buflen < 2) {
+				errno = ENOMEM;
+				return -1;
+			}
+			*(*ctx->buf)++ = ctx->code;
+			ctx->len = (*ctx->buf)++;
+			*ctx->len = 0;
+			r += 2;
```
billie-alsup, line 755: Do we not risk a buffer overrun because we do not adjust ctx->buflen by the two bytes we just inserted?  We appear to not account for those two header bytes on each invocation (it seems the same context can be passed in sequential invocations of rfc3396_write), or even if there is a large write, we don't account for those two bytes in each iteration of the loop.

src/dhcp.c
```diff
@@ -723,6 +723,76 @@ dhcp_message_add_addr(struct bootp *bootp,
...
+		wlen = len < RFC3396_BOUNDARY ? len : RFC3396_BOUNDARY;
+		left = RFC3396_BOUNDARY - *ctx->len;
+		if (left < wlen)
+			wlen = left;
+		if (ctx->buflen < wlen) {
+			errno = ENOMEM;
+			return -1;
+		}
+
+		memcpy(*ctx->buf, datap, wlen);
```
billie-alsup, line 767: The result being that it is possible to have a buffer overrun here.

src/dhcp.c
rsmarples: We do and this should be fixed in 0f303ab4

PR #384, opened 2024-10-11
Apply lastlease behavior to DHCPv6

jvfranklin:
If no response is received to Confirm messages, instead of failing, go ahead and use the lease, as recommended by RFC 8415 18.2.3.


rsmarples:
Thanks!

PR #387, opened 2024-10-17
DHCP6: lastlease behavior after Confim non-response

jvfranklin:
If lastlease is enabled, and dhcpcd is unable to confirm its prior lease, after timeout, don't continue to solicit. Confine lastlease behavior to only the CONFIRM state.

This cleans up some minor issues with the initial commit for DHCPv6 lastlease.

### DIFF
--- a/src/dhcp6.c
+++ b/src/dhcp6.c
@@ -187,6 +187,7 @@ static const char * const dhcp6_statuses[] = {
 
 static void dhcp6_bind(struct interface *, const char *, const char *);
 static void dhcp6_failinform(void *);
+static void dhcp6_startrebind(void *arg);
 static void dhcp6_recvaddr(void *, unsigned short);
 static void dhcp6_startdecline(struct interface *);
 static void dhcp6_startrequest(struct interface *);
@@ -1760,8 +1761,12 @@ dhcp6_fail(struct interface *ifp, bool drop)
 		dhcp_unlink(ifp->ctx, state->leasefile);
 		dhcp6_addrequestedaddrs(ifp);
 		eloop_timeout_delete(ifp->ctx->eloop, NULL, ifp);
-	} else if (state->recv && ifp->options->options & DHCPCD_LASTLEASE) {
+	} else if ((state->state == DH6S_CONFIRM || state->state == DH6S_REBIND) &&
+	           ifp->options->options & DHCPCD_LASTLEASE) {
 		dhcp6_bind(ifp, NULL, NULL);
+		state->state = DH6S_REBIND;
+		dhcp6_startrebind(ifp);
+		return;
 	} else if (state->new) {
 		script_runreason(ifp, "TIMEOUT6");
 		// We need to keep the expire timeout alive


src/dhcp6.c
```diff
@@ -1760,8 +1760,9 @@ dhcp6_fail(struct interface *ifp, bool drop)
 		dhcp_unlink(ifp->ctx, state->leasefile);
 		dhcp6_addrequestedaddrs(ifp);
 		eloop_timeout_delete(ifp->ctx->eloop, NULL, ifp);
-	} else if (state->recv && ifp->options->options & DHCPCD_LASTLEASE) {
+	} else if (state->state == DH6S_CONFIRM && ifp->options->options & DHCPCD_LASTLEASE) {
 		dhcp6_bind(ifp, NULL, NULL);
+		return;
```
rsmarples, line 1765: Why return?

We couldn't find any DHCPv6 server, so we do use the last lease, but we should retry and DISCOVER yes?

src/dhcp6.c
```diff
@@ -1760,8 +1760,9 @@ dhcp6_fail(struct interface *ifp, bool drop)
 		dhcp_unlink(ifp->ctx, state->leasefile);
 		dhcp6_addrequestedaddrs(ifp);
 		eloop_timeout_delete(ifp->ctx->eloop, NULL, ifp);
-	} else if (state->recv && ifp->options->options & DHCPCD_LASTLEASE) {
+	} else if (state->state == DH6S_CONFIRM && ifp->options->options & DHCPCD_LASTLEASE) {
```
rsmarples, line 1763: Can we also allow the state DH6S_REBIND please?
IF you ask for a Prefix Delegation you REBIND, otherwise CONFIRM and ideally we want the behavior in both states.
And we should still check state->recv to ensure we ARE using the last lease in this state.

jvfranklin: Sure, I'm not seeing any issues with those changes in my own testing.

src/dhcp6.c
jvfranklin: Soliciting for a lease while using the current one could lead to problems if the DHCPv6 server has ping checking enabled, in the case where you come back up on the same network. We've seen it fail the ping check (ie, gets a response) and issue a new IP address when that happens, thinking there's an address conflict.  I think making this behavior optional with the lastlease option allows everybody to choose which way they want to go, since it is kind of a couple of competing requirements we're trying to untangle here. What do you think?

rsmarples: Does it matter it gets a new IP address? I mean, we have just failed to CONFIRM it?
This behavior is long standing in the DHCPv4 side of things and am just trying to get the same behavior both sides.

jvfranklin: For the product I work on, it would for sure. We've had customer complaints due to address changes like this in the past. It wouldn't fly for us at all.

rsmarples: Well, DHCP address can and do change. Shouldn't count on them being static. Even if you get a new address, the old one will still persist here.

Anyway, how would you feel instead with moving to the REBIND state immediately then?

jvfranklin: That sounds like a very good idea. Let me try it out against the Ready Logo tests to make sure it doesn't cause any unexpected behaviors, though I don't really expect it to, and I'll update this PR.

Is this the right way to do it?
dhcp6_fail()
} else if (state->state == DH6S_CONFIRM && ifp->options->options & DHCPCD_LASTLEASE) {
    dhcp6_bind(ifp, NULL, NULL);
    state->state = DH6S_REBIND;
    dhcp6_startrebind(ifp);
    return;
}

rsmarples: Looks ok here.

src/dhcp6.c
rsmarples: I still feel we should check if we have recv still. Why did you remove it?

jvfranklin: I've got it in my local branch, just haven't pushed it yet. It'll be correct in my next push along with the rebind change.

jvfranklin: Thanks! I've updated the commit now.

PR #389, opened 2024-10-23
Allow limited RLIMIT_FSIZE when dumping lease

### LINKED ISSUE #386: dumplease redirect stdout blocked by rlimit (SIGXFSZ)
dumping lease to stdout works fine, but if I try to redirect it, rlimit kicks in and causes exit with file size signal.
```
dhcpcd --config /etc/dhcp/dhcpcd.conf --dumplease -4 eth0 > lease.txt
Program terminated with signal SIGXFSZ, File size limit exceeded.
```
if PRIVSEP but ctx->options is DHCPCD_DUMPLEASE, we should relax the rlimit to a reasonable lease file size while dumping the lease.

ColinMcInnes:
When PRIVSEP, set rlimit to 5Kb when dumping lease (all other privileges are still dropped).

Without this, redirecting dumplease to anything but stdout results in SIGXFSZ due to EBIG error.

Resolves #386

### DIFF
--- a/src/dhcpcd.c
+++ b/src/dhcpcd.c
@@ -1739,6 +1739,13 @@ dhcpcd_readdump2(void *arg, unsigned short events)
 	if (ctx->ctl_buf[ctx->ctl_buflen - 1] != '\0') /* unlikely */
 		ctx->ctl_buf[ctx->ctl_buflen - 1] = '\0';
 	script_dump(ctx->ctl_buf, ctx->ctl_buflen);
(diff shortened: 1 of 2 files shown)


ColinMcInnes:
Note this will still not let you dump lease as part of an exit hook script, those are still limited to FSIZE 0.

ColinMcInnes:
> @ColinMcInnes I made the fix a bit simpler. Does it still work for you?

Yes that works for me.

PR #397, opened 2024-11-01
DHCP6: IPv6 Ready Logo test fixes

jvfranklin:
Implement USE_MULTICAST
Handle server-directed SOL_MAX_RT and INF_MAX_RT values Perform per RFC 8415 for NotOnLink and NoBinding error conditions Don't attempt to renew expired leases
Continue sending Info-Request messages until response is received
MAX_RC values aligned with RFC definitions

### DIFF
--- a/src/dhcpcd-definitions-small.conf
+++ b/src/dhcpcd-definitions-small.conf
@@ -15,7 +15,7 @@ define 12	dname			host_name
 define 15	array dname		domain_name
 define 26	uint16			interface_mtu
 define 28	request ipaddress	broadcast_address
-define 33	request array ipaddress	static_routes
+define 33	array ipaddress		static_routes
 define 50	ipaddress		dhcp_requested_address
 define 51	uint32			dhcp_lease_time
 define 52	byte			dhcp_option_overload
--- a/src/dhcpcd-definitions.conf
+++ b/src/dhcpcd-definitions.conf
@@ -45,7 +45,7 @@ define 29	byte			perform_mask_discovery
 define 30	byte			mask_supplier
 define 31	byte			router_discovery
 define 32	ipaddress		router_solicitation_address
-define 33	request array ipaddress	static_routes
+define 33	array ipaddress		static_routes
 define 34	byte			trailer_encapsulation
 define 35	uint32			arp_cache_timeout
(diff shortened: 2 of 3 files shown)


src/dhcp6.c
```diff
@@ -1714,14 +1721,8 @@ dhcp6_startinform(void *arg)
...
-	dhcp6_sendinform(ifp);
-	/* RFC3315 18.1.2 says that if CONFIRM failed then the prior addresses
-	 * SHOULD be used. The wording here is poor, because the addresses are
-	 * merely one facet of the lease as a whole.
-	 * This poor wording might explain the lack of similar text for INFORM
-	 * in 18.1.5 because there are no addresses in the INFORM message. */
-	eloop_timeout_add_sec(ifp->ctx->eloop,
-	    INF_MAX_RD, dhcp6_failinform, ifp);
+	else
+		dhcp6_sendinform(ifp);
```
rsmarples, line 1725: OK, I see what you're trying to fix here. We want to drop through to failinform so we can use the lastlease code, but only on the first failure.

So I think what we need to do is after we have logged the message we want to guard the state timing changes like so

```
if (state->state == DH6S_INFORM) {
      dhcp6_sendinform(ifp);
      return;
}
```

Then we change failinform to do dhcp6_fail(ifp, false) - ie we don't force a drop.
Adjust the code there to handle lastlease for INFORM and we should be good yes?

src/dhcp6.c
```diff
@@ -3439,16 +3486,20 @@ dhcp6_recvif(struct interface *ifp, const char *sfrom,
...
+				if ((state->state == DH6S_RENEW ||
+				     state->state == DH6S_REBIND) &&
+				     state->has_no_binding) {
+					dhcp6_startrequest(ifp);
+				}
+				else if (state->state == DH6S_REQUEST &&
+				         state->lerror == D6_STATUS_NOTONLINK) {
+					eloop_timeout_delete(ifp->ctx->eloop, NULL, ifp);
+					dhcp6_startdiscover(ifp);
+				}
```
rsmarples, line 3502: We are no longer handling the state where we got a reply without an error but it didn't contain any addresses with error codes either.

rsmarples:
@jvfranklin I've merged some of your changes here into master now.

PR #408, opened 2024-11-21
Fix using multiple enterprise IDs with vendclass (Option 124 DHCP / Option 16 DHCPv6) (#328)

spoljak-ent:
As it's outlined in #328, using multiple enterprise IDs with vendclass overwrites all previous values.
This fixes the malformation of the DHCPv4 packet and adds support for multiple different enterprise ID vendclass options.

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -808,7 +808,6 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 	const struct dhcp_lease *lease = &state->lease;
 	char hbuf[HOSTNAME_MAX_LEN + 1];
 	const char *hostname;
-	const struct vivco *vivco, *vivco_endp = ifo->vivco + ifo->vivco_len;;
 	int mtu;
 #ifdef AUTH
 	uint8_t *auth, auth_len;
@@ -1137,38 +1136,35 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
 		       p += ifo->mudurl[0] + 1;
 		}
 
+#ifndef SMALL
 		if (ifo->vivco_len &&
 		    !has_option_mask(ifo->nomask, DHO_VIVCO))
 		{
-			AREA_CHECK(sizeof(ul));
-			*p++ = DHO_VIVCO;
-			size_t totallen = 0;
-			uint8_t datalen, datalenopt;
-			for (vivco = ifo->vivco; vivco != vivco_endp; vivco++)
-				totallen += sizeof(uint32_t) + 2 * sizeof(uint8_t) + vivco->len;
-			if (totallen > UINT8_MAX) {
-				logerrx("%s: VIVCO option too big",
-					ifp->name);
-				free(bootp);
-				return -1;
-			}
-			*p++ = (uint8_t)totallen;
-			for (vivco = ifo->vivco; vivco != vivco_endp; vivco++) {
+			struct vivco *vivco = ifo->vivco;
+			size_t vlen = ifo->vivco_len;
+			struct rfc3396_ctx rctx = {
+				.code = DHO_VIVCO,
+				.buf = &p,
+				.buflen = AREA_LEFT,
+			};
+
+			for (; vlen > 0; vivco++, vlen--) {
 				ul = htonl(vivco->en);
-				memcpy(p, &ul, sizeof(ul));
-				p += sizeof(ul);
-				datalen = (uint8_t)(sizeof(uint8_t) + vivco->len);
-				memcpy(p, &datalen, sizeof(datalen));
-				p += sizeof(datalen);
-				datalenopt = (uint8_t)(vivco->len);
-				memcpy(p, &datalenopt, sizeof(datalenopt));
-				p += sizeof(datalenopt);
-				memcpy(p, vivco->data, vivco->len);
-				p += vivco->len;
+				if (rfc3396_write(&rctx, &ul, sizeof(ul)) == -1)
+					goto toobig;
+				lp = rfc3396_zero(&rctx);
+				if (lp == NULL)
+					goto toobig;
+				if (rfc3396_write_byte(&rctx,
+					(uint8_t)vivco->len) == -1)
+					goto toobig;
+				if (rfc3396_write(&rctx,
+					vivco->data, vivco->len) == -1)
+					goto toobig;
+				*lp = (uint8_t)(*lp + vivco->len + 1);
 			}
 		}
-
-#ifndef SMALL
+		
 		if (ifo->vsio_len &&
 		    !has_option_mask(ifo->nomask, DHO_VIVSO))
 		{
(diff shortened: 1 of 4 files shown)


src/if-options.c
```diff
@@ -2119,6 +2120,12 @@ parse_option(struct dhcpcd_ctx *ctx, const char *ifname, struct if_options *ifo,
 			logerrx("invalid code: %s", arg);
 			return -1;
 		}
+		for (vivco = ifo->vivco; vivco != vivco_endp; vivco++) {
+			if (vivco->en == (uint32_t)u) {
+				logerrx("only one vendor class option per enterprise number");
+				return -1;
```
rsmarples, line 2126: Maybe a better text would be "vendor class option for enterprise number %u already defined".
We might also want to consider do we want to have a different option sent for the same enterprise number based on interface or profile name? If so, we might just want to silently override it instead.

Thoughts?

src/dhcp.c
```diff
@@ -1142,26 +1142,29 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
...
-				if (vivco->len + 2 + *lp > 255) {
-					logerrx("%s: VIVCO option too big",
-					    ifp->name);
-					free(bootp);
-					return -1;
-				}
-				*p++ = (uint8_t)vivco->len;
+			size_t totallen = 0;
+			uint8_t datalen, datalenopt;
+			for (vivco = ifo->vivco; vivco != vivco_endp; vivco++)
```
rsmarples, line 1147: White space between the declaration and the code please, but see below comment first.

src/dhcp.c
```diff
@@ -1142,26 +1142,29 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
...
+				ul = htonl(vivco->en);
+				memcpy(p, &ul, sizeof(ul));
+				p += sizeof(ul);
+				datalen = (uint8_t)(sizeof(uint8_t) + vivco->len);
+				memcpy(p, &datalen, sizeof(datalen));
+				p += sizeof(datalen);
+				datalenopt = (uint8_t)(vivco->len);
+				memcpy(p, &datalenopt, sizeof(datalenopt));
+				p += sizeof(datalenopt);
 				memcpy(p, vivco->data, vivco->len);
```
rsmarples, line 1166: This could be re-written as
```
*p++ = (uint8_t)(sizeof(uint8_t) + vivco->len);
*p++ = (uint8_t)(vivco->len);
```

Much smaller code and no need to declare the variables then.

rsmarples (changes requested):

rsmarples:
RFC 3925 says
> Multiple instances of this option may be present
   and MUST be concatenated in accordance with [RFC 3396](https://datatracker.ietf.org/doc/html/rfc3396) [[4](https://datatracker.ietf.org/doc/html/rfc3925#ref-4)].

So we should try and use the rfc3396 encoding functions we added in the previous PR to help here.

src/if-options.c
spoljak-ent: Yeah, it is better regarding the text.
Regarding the different option, I was going along with RFC 3925 that we should have only one enterprise number among all instances of this option.
How would this work with the silent override?

src/dhcp.c
spoljak-ent: Agreed, I'll fix it.

spoljak-ent:
> RFC 3925 says
> 
> > Multiple instances of this option may be present
> > and MUST be concatenated in accordance with [RFC 3396](https://datatracker.ietf.org/doc/html/rfc3396) [[4](https://datatracker.ietf.org/doc/html/rfc3925#ref-4)].
> 
> So we should try and use the rfc3396 encoding functions we added in the previous PR to help here.

We could, but I wasn't sure about adding (much) more code since the previous PR was wrapped in #ifndef SMALL.

rsmarples:
> > RFC 3925 says
> > > Multiple instances of this option may be present
> > > and MUST be concatenated in accordance with [RFC 3396](https://datatracker.ietf.org/doc/html/rfc3396) [[4](https://datatracker.ietf.org/doc/html/rfc3925#ref-4)].
> > 
> > 
> > So we should try and use the rfc3396 encoding functions we added in the previous PR to help here.
> 
> We could, but I wasn't sure about adding (much) more code since the previous PR was wrapped in #ifndef SMALL.

We should wrap this code in `#ifndef SMALL` as well really.

spoljak-ent:
> > > RFC 3925 says
> > > > Multiple instances of this option may be present
> > > > and MUST be concatenated in accordance with [RFC 3396](https://datatracker.ietf.org/doc/html/rfc3396) [[4](https://datatracker.ietf.org/doc/html/rfc3925#ref-4)].
> > > 
> > > 
> > > So we should try and use the rfc3396 encoding functions we added in the previous PR to help here.
> > 
> > 
> > We could, but I wasn't sure about adding (much) more code since the previous PR was wrapped in #ifndef SMALL.
> 
> We should wrap this code in `#ifndef SMALL` as well really.

Got it. I'll make the changes.

src/if-options.c
spoljak-ent: Sorry for the confusion, I meant about having a different option sent for the same enterprise number based on interface or profile name. Haven't worked with multiple profiles/interfaces tbh.

src/dhcp.c
spoljak-ent: Made the requested change with RFC3396 ctx. 
Looks much cleaner now.

src/dhcp.c
spoljak-ent: Resolved with RFC3396 ctx.

spoljak-ent:
> > > > RFC 3925 says
> > > > > Multiple instances of this option may be present
> > > > > and MUST be concatenated in accordance with [RFC 3396](https://datatracker.ietf.org/doc/html/rfc3396) [[4](https://datatracker.ietf.org/doc/html/rfc3925#ref-4)].
> > > > 
> > > > 
> > > > So we should try and use the rfc3396 encoding functions we added in the previous PR to help here.
> > > 
> > > 
> > > We could, but I wasn't sure about adding (much) more code since the previous PR was wrapped in #ifndef SMALL.
> > 
> > 
> > We should wrap this code in `#ifndef SMALL` as well really.
> 
> Got it. I'll make the changes.

All required changes done... for now 😄

src/dhcp.c
```diff
@@ -1137,35 +1136,35 @@ make_message(struct bootp **bootpm, const struct interface *ifp, uint8_t type)
...
+
+			for (; vlen > 0; vivco++, vlen--) {
+				ul = htonl(vivco->en);
+				if (rfc3396_write(&rctx, &ul, sizeof(ul)) == -1)
+					goto toobig;
+				lp = rfc3396_zero(&rctx);
+				if (lp == NULL)
+					goto toobig;
+				if (rfc3396_write_byte(&rctx,
+					(uint8_t)vivco->len) == -1)
```
rsmarples, line 1159: Where the line breaks or flows onto the next, please use 4 spaces rather than a tab.
I should really force an opinionated formatter on pull requests at some point :)

rsmarples (commented):
This is looking good now, thanks.

I'm currently in the process of moving house so won't get a chance to actually test it for Some Time.

src/dhcp.c
spoljak-ent: Noted and corrected :)

perkelix:
@rsmarples ready to merge?

Issue #428, opened 2025-01-06
default RA lifetime not set in route


ColinMcInnes:
the default route is being sent by the router via RA, and they are setting a lifetime of 1800s and a hop limit of 64. However this is not showing up in the ip -6 route show. 

```
$ ip -6 route show ::/0
default via fe80::2efa:a2ff:fe90:e275 dev eth0 proto ra metric 1002 pref medium

$ dhcpcd -U eth0
reason=ROUTERADVERT
interface=eth0
protocol=ra
nd1_from=fe80::2efa:a2ff:fe90:e275
nd1_acquired=1100
nd1_now=1152
nd1_hoplimit=64
nd1_flags=M
nd1_lifetime=1800
nd1_source_address=2cfaa290e275
nd1_prefix_information1_length=64
nd1_prefix_information1_flags=L
nd1_prefix_information1_vltime=2592000
nd1_prefix_information1_pltime=604800
nd1_prefix_information1_prefix=2001:192:168:228::
```
dhcpcd 10.1.0 linux

here's what it looks like if I use ISC's dhclient 4.4.3
```
default via fe80::2efa:a2ff:fe90:e275 dev eth0 proto ra metric 1024 expires 1609sec hoplimit 64 pref medium
```

Any ideas where to start looking?

ColinMcInnes:
Found it, in if-linux and ipv6. I'll create a patch to send it to netlink.

rsmarples:
ISC dhclient doesn't set any routes, just addresses.
Something Else (TM) must be setting the routes :)

dhcpcd will expire the routes anyway, so any fix we do is purely cosmetic (unless of course dhcpcd has stopped for whatever reason)

ColinMcInnes:
> dhcpcd will expire the routes anyway, so any fix we do is purely cosmetic (unless of course dhcpcd has stopped for whatever reason)

Yes, I agree it's purely cosmetic. I have some other code that reports the current route lifetime from nl, so this is why I'm adding it for informational purposes.

> ISC dhclient doesn't set any routes, just addresses. Something Else (TM) must be setting the routes :)

I found that out once I started looking deeper. :)

juw:
Unfortunately the change to add the lifetime to the route is not only cosmetic. It seems, that the lifetime does not get updated once a new RA is received and hence the route will expire, removed and only then re-added once a new RA is received. While before this did not happen.

ColinMcInnes:
[...]

PR #429, opened 2025-01-08
Add route lifetime from Router Advertisement

ColinMcInnes:
Sends route lifetime to NETLINK RTA_EXPIRES

Partially addresses #428

Still need to send hoplimit

### DIFF
--- a/src/common.c
+++ b/src/common.c
@@ -219,13 +219,14 @@ is_root_local(void)
 uint32_t
 lifetime_left(uint32_t lifetime, const struct timespec *acquired, const struct timespec *now)
 {
-    uint32_t elapsed;
+	uint32_t elapsed;
 
-    if (lifetime == INFINITE_LIFETIME)
-        return lifetime;
+	if (lifetime == INFINITE_LIFETIME)
+		return lifetime;
 
-    elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
-    if (elapsed > lifetime)
-        return 0;
-    return lifetime - elapsed;
+	elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
+	if (elapsed > lifetime)
+		return 0;
+
+	return lifetime - elapsed;
 }
--- a/src/common.c
+++ b/src/common.c
@@ -224,7 +224,7 @@ lifetime_left(uint32_t lifetime, const struct timespec *acquired, const struct t
 	if (lifetime == INFINITE_LIFETIME)
 		return lifetime;
 
-	elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
+	elapsed = (uint32_t)eloop_timespec_diff(now, acquired, NULL);
 	if (elapsed > lifetime)
 		return 0;
 
--- a/src/common.c
+++ b/src/common.c
@@ -214,3 +214,17 @@ is_root_local(void)
 	return -1;
 #endif
 }
+
+uint32_t
+lifetime_left(uint32_t lifetime, const struct timespec *aquired, const struct timespec *now)
+{
+    uint32_t elpased;
+
+    if (lifetime == INFINITE_LIFETIME)
+        return lifetime;
+
+    elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
+    if (elapsed > lifetime)
+        return 0;
+    return lifetime - elapsed;
+}
--- a/src/common.c
+++ b/src/common.c
@@ -38,6 +38,7 @@
 
 #include "common.h"
 #include "dhcpcd.h"
+#include "eloop.h"
 #include "if-options.h"
 
 const char *
@@ -216,9 +217,9 @@ is_root_local(void)
 }
 
 uint32_t
-lifetime_left(uint32_t lifetime, const struct timespec *aquired, const struct timespec *now)
+lifetime_left(uint32_t lifetime, const struct timespec *acquired, const struct timespec *now)
 {
-    uint32_t elpased;
+    uint32_t elapsed;
 
     if (lifetime == INFINITE_LIFETIME)
         return lifetime;
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -753,7 +753,7 @@ if_copyrt(struct dhcpcd_ctx *ctx, struct rt *rt, struct nlmsghdr *nlm)
 			break;
 		}
 		case RTA_EXPIRES:
-			rt->rt_expires = *(unsigned int *)RTA_DATA(rta);
+			rt->rt_expires = *(uint32_t *)RTA_DATA(rta);
 			break;
 		}
 
@@ -1739,7 +1739,6 @@ if_route(unsigned char cmd, const struct rt *rt)
 		add_attr_32(&nlm.hdr, sizeof(nlm), RTA_OIF, rt->rt_ifp->index);
 
 	/* add route lifetime */
-	logdebugx("Sending RTA_EXPIRES %d", rt->rt_expires);
 	add_attr_32(&nlm.hdr, sizeof(nlm), RTA_EXPIRES, rt->rt_expires);
 
 	if (rt->rt_metric != 0)
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -754,10 +754,7 @@ if_copyrt(struct dhcpcd_ctx *ctx, struct rt *rt, struct nlmsghdr *nlm)
 		}
 		case RTA_EXPIRES:
 		{
-			if (rt->rt_expires != 0)
-			{
-				rt->rt_expires = *(uint32_t *)RTA_DATA(rta);
-			}
+			rt->rt_expires = *(uint32_t *)RTA_DATA(rta);
 			break;
 		}
 		}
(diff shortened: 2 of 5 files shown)


rsmarples:
This looks nice! Can we also extract the lifetime in if_copyrt in if-linux.c please?

ColinMcInnes:
> This looks nice! Can we also extract the lifetime in if_copyrt in if-linux.c please?

Yes, I can do that.

rsmarples (changes requested):

src/if-linux.c
```diff
@@ -1735,6 +1738,10 @@ if_route(unsigned char cmd, const struct rt *rt)
 	if (!sa_is_loopback(&rt->rt_gateway))
 		add_attr_32(&nlm.hdr, sizeof(nlm), RTA_OIF, rt->rt_ifp->index);
 
+	/* add route lifetime */
+	logdebugx("Sending RTA_EXPIRES %d", rt->rt_expires);
```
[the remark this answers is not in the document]
ColinMcInnes: done

src/route.h
```diff
@@ -120,6 +120,7 @@ struct rt {
 #define	RTDF_GATELINK		0x40		/* Gateway is on link */
 	size_t			rt_order;
 	rb_node_t		rt_tree;
+	int			rt_expires;	/* current lifetime of route */
```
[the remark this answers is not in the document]
ColinMcInnes: done

src/if-linux.c
```diff
@@ -752,6 +752,9 @@ if_copyrt(struct dhcpcd_ctx *ctx, struct rt *rt, struct nlmsghdr *nlm)
 			}
 			break;
 		}
+		case RTA_EXPIRES:
+			rt->rt_expires = *(unsigned int *)RTA_DATA(rta);
```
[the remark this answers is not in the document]
ColinMcInnes: From some documents it looks like NETLINK outputs this as an int, because it's possible for a route to be expired and the lifetime moves into the negative. Would this handle that case when copying a route?
https://man7.org/linux/man-pages/man7/rtnetlink.7.html

src/ipv6.c
```diff
@@ -2370,6 +2370,8 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 #ifdef HAVE_ROUTE_PREF
 		rt->rt_pref = ipv6nd_rtpref(rap->flags);
 #endif
+		rt->rt_expires = rap->lifetime;
```
rsmarples, line 2373: OK, this won't work. We need to work out elapsed seconds from acquisition for the real expiry.

Top of function: 
```
struct timespec now;
uint32_t elapsed;

clock_gettime(CLOCK_MONOTONIC, &now);
```
For each route:
```
elapsed = (uint32_t)eloop_timespec_diff(&now, &rap->acquired, NULL);
rt->rt_expires = rap->lifetime - elapsed;
```

If you scan up the code you can do the similar with rinfo and addrs so those routes get expiry times as well.

ColinMcInnes: Will do

src/if-linux.c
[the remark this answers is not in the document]
rsmarples: I'm looking at the kernel code where it's declared as NLA_U32.
https://elixir.bootlin.com/linux/v6.12.6/source/net/ipv6/route.c#L5000

I'm guessing the fine man page here is wrong.
If a route expires then it would be removed like an address?

src/ipv6.c
```diff
@@ -2301,6 +2301,10 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 	const struct routeinfo *rinfo;
 	const struct ipv6_addr *addr;
 	struct in6_addr netmask;
+	struct timespec now;
+	uint32_t elapsed;
+
+	clock_gettime(CLOCK_MONOTONIC, &now);
```
rsmarples, line 2307: Could move this underneath the below check please? No need for the syscall if we don't have any routers.

src/ipv6.c
```diff
@@ -2325,6 +2329,8 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 #ifdef HAVE_ROUTE_PREF
 			rt->rt_pref = ipv6nd_rtpref(rinfo->flags);
 #endif
+			elapsed = (uint32_t)eloop_timespec_diff(&now, &rap->acquired, NULL);
```
rsmarples, line 2332: Should be &rinfo->acquired, not rap. Sorry, my fault not being clear.

src/ipv6.c
```diff
@@ -2339,6 +2345,9 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 #ifdef HAVE_ROUTE_PREF
 				rt->rt_pref = ipv6nd_rtpref(rap->flags);
 #endif
+				elapsed = (uint32_t)eloop_timespec_diff(&now, &rap->acquired, NULL);
```
rsmarples, line 2348: Should be &addr->acquired, not rap. Sorry, my fault not being clear.

src/if-linux.c
```diff
@@ -1739,7 +1739,6 @@ if_route(unsigned char cmd, const struct rt *rt)
 		add_attr_32(&nlm.hdr, sizeof(nlm), RTA_OIF, rt->rt_ifp->index);
 
 	/* add route lifetime */
-	logdebugx("Sending RTA_EXPIRES %d", rt->rt_expires);
 	add_attr_32(&nlm.hdr, sizeof(nlm), RTA_EXPIRES, rt->rt_expires);
```
rsmarples, line 1742: I think we should only add this if we have a non zero rt_expires value.

rsmarples (changes requested):
Gah, and I forgot one last thing.

Before setting each rt_expires, check that the lifetime (before working out elapsed) is not ND6_INFINITE_LIFETIME.

ColinMcInnes:
> ND6_INFINITE_LIFETIME

Something like this?

```
		if (rap->lifetime != ND6_INFINITE_LIFETIME)
		{
			elapsed = (uint32_t)eloop_timespec_diff(&now, &rap->acquired, NULL);
			rt->rt_expires = rap->lifetime - elapsed;
		}
		else
			rt->rt_expires = rap->lifetime;
```

rsmarples:
Exactly! We have a lot of that code, so maybe we benefit from a helper function in common.c?

```
uint32_t
lifetime_left(uint32_t lifetime, const struct timespec *aquired, const struct timespec *now)
{
    uint32_t elpased;

    if (lifetime == INFINITE_LIFETIME)
        return lifetime;

    elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
    if (elapsed > lifetime)
        return 0;
    return lifetime - elapsed;
}
```

then we can simply write:

`rt->rt_expires = lifetime_left(rap->lifetime, &rap->aquired, &now);`

Thoughts? If you like it, `#define INFINITE_LIFETIME (~0U)` please in common.h
A future patch could then replace bits where we do this to reduce code even more.

src/common.c
```diff
@@ -214,3 +215,17 @@ is_root_local(void)
...
+    uint32_t elapsed;
+
+    if (lifetime == INFINITE_LIFETIME)
+        return lifetime;
+
+    elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
+    if (elapsed > lifetime)
+        return 0;
+    return lifetime - elapsed;
+}
```
rsmarples, line 231: Please use tabs and not spaces to indent with.

This was my bad as I don't know how to insert a space in a browser!

src/ipv6.c
```diff
@@ -2325,6 +2328,7 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 #ifdef HAVE_ROUTE_PREF
 			rt->rt_pref = ipv6nd_rtpref(rinfo->flags);
 #endif
+			rt->rt_expires = lifetime_left(rap->lifetime, &rinfo->acquired, &now);
```
rsmarples, line 2331: Please use rinfo->lifetime.

src/ipv6.c
```diff
@@ -2339,6 +2343,8 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 #ifdef HAVE_ROUTE_PREF
 				rt->rt_pref = ipv6nd_rtpref(rap->flags);
 #endif
+				rt->rt_expires = lifetime_left(rap->lifetime, &addr->acquired, &now);
```
rsmarples, line 2346: Please use addr->prefix_vltime.

rsmarples (changes requested):
Last round of changes, then I think we're good to merge!

src/common.c
```diff
@@ -214,3 +215,18 @@ is_root_local(void)
...
+
+uint32_t
+lifetime_left(uint32_t lifetime, const struct timespec *acquired, const struct timespec *now)
+{
+	uint32_t elapsed;
+
+	if (lifetime == INFINITE_LIFETIME)
+		return lifetime;
+
+	elapsed = (uint32_t)eloop_timespec_diff(now, &acquired, NULL);
```
rsmarples, line 227: We want to pass acquired by value it’s already a pointer.

src/ipv6.c
```diff
@@ -2305,6 +2305,9 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 	if (ctx->ra_routers == NULL)
 		return 0;
 
+	struct timespec now;
```
rsmarples, line 2308: Please move this declaration to the top section.

rsmarples (changes requested):
The CI picked up a warning we need to address, but this is looking really good now.

rsmarples:
@ColinMcInnes thanks for the excellent work here.

rsmarples:
@ColinMcInnes I used the new function to reduce more code in 7b94a2e6e0aa5c147ccda2ad3b09931ed77fb6ff

Hopefully everything still works!

PR #441, opened 2025-01-24
Update route if acquired time changes

### LINKED ISSUE #428: default RA lifetime not set in route

ColinMcInnes:
If a route gets an update (say from an RA), consider it changed so new expiry gets pushed out to system.

Resolves #428

### DIFF
--- a/src/route.c
+++ b/src/route.c
@@ -556,7 +556,7 @@ rt_add(rb_tree_t *kroutes, struct rt *nrt, struct rt *ort)
 	    sa_cmp(&ort->rt_gateway, &nrt->rt_gateway) == 0)
 	{
 		/* If it has not been renewed by RA, and MTU is unchanged, skip */
-		if (ort->rt_updated == nrt->rt_updated &&
+		if (ort->rt_updated.tv_sec == nrt->rt_updated.tv_sec &&
 			ort->rt_mtu == nrt->rt_mtu)
 			return true;
 		change = true;
@@ -681,7 +681,7 @@ rt_doroute(rb_tree_t *kroutes, struct rt *rt)
 		    (rt->rt_ifa.sa_family != AF_UNSPEC &&
 		    sa_cmp(&or->rt_ifa, &rt->rt_ifa) != 0) ||
 		    or->rt_mtu != rt->rt_mtu ||
-			or->rt_updated != rt->rt_updated)
+			or->rt_updated.tv_sec != rt->rt_updated.tv_sec)
 		{
 			if (!rt_add(kroutes, rt, or))
 				return false;
--- a/src/route.c
+++ b/src/route.c
@@ -502,22 +502,30 @@ rt_recvrt(int cmd, const struct rt *rt, pid_t pid)
 #endif
 }
 
-/* Compare route details */
+/* Compare miscellaneous route details */
 static bool
-rt_changed(struct rt *nrt, struct rt *ort)
+rt_cmp_misc(struct rt *nrt, struct rt *ort)
 {
 	/* MTU changed */
 	if (ort->rt_mtu != nrt->rt_mtu)
-		return true;
+		return false;
 
 #ifdef HAVE_ROUTE_LIFETIME
-	/* If the expired target time is different, we have updated our expired time
-	 * from a new RA and should pass that value on to the system */
-	if (abs((int64_t)nrt->rt_expires - (int64_t)ort->rt_expires) > 30)
-		return true;
+	uint32_t deviation;
+
+	/* There might be a minor difference between kernel route
+	 * lifetime and our lifetime due to processing times.
+	 * We allow a small deviation to avoid needless route changes.
+	 * dhcpcd will expire the route regardless of route lifetime support. */
+	if (nrt->rt_lifetime > ort->rt_lifetime)
+		deviation = nrt->rt_lifetime - ort->rt_lifetime;
+	else
+		deviation = ort->rt_lifetime - nrt->rt_lifetime;
+	if (deviation > RTLIFETIME_DEV_MAX)
+		return false;
 #endif
 
-	return false;
+	return true;
 }
 
 static bool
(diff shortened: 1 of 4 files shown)


src/route.c
```diff
@@ -555,7 +555,9 @@ rt_add(rb_tree_t *kroutes, struct rt *nrt, struct rt *ort)
 	    rt_cmp_netmask(ort, nrt) == 0 &&
 	    sa_cmp(&ort->rt_gateway, &nrt->rt_gateway) == 0)
 	{
-		if (ort->rt_mtu == nrt->rt_mtu)
+		/* If it has not been renewed by RA, and MTU is unchanged, skip */
+		if (ort->rt_updated.tv_sec == nrt->rt_updated.tv_sec &&
+			ort->rt_mtu == nrt->rt_mtu)
```
rsmarples, line 560: Can we not just test if rt_expires is different?

ColinMcInnes: The code originally checked if mtu was identical, we still want it to skip the push if mtu is unchanged, right?

rsmarples: I want to skip the push if the route is unchanged. So compare mtu and expires - if the same we skip.

ColinMcInnes:
I have tested this fix with a 600s RA renewal with 1800s lifetime. Every 10 minutes or so we see the RA come in, and I see the route lifetime get re-updated to 1800s each time.

src/route.h
```diff
@@ -121,6 +121,7 @@ struct rt {
 	size_t			rt_order;
 	rb_node_t		rt_tree;
 	uint32_t		rt_expires;	/* current lifetime of route */
+	struct timespec	rt_updated; /* time route was last updated */
```
rsmarples, line 124: This should be rt_aquired as that's the time it uses. Updated implies it's updated when the route is added which isn't the case. While this may not be visible on Linux as it handles the same route per interface other OS's like the BSD's don't.

src/route.c
```diff
@@ -540,9 +542,12 @@ rt_add(rb_tree_t *kroutes, struct rt *nrt, struct rt *ort)
 #endif
 		    sa_cmp(&ort->rt_gateway, &nrt->rt_gateway) == 0)))
 		{
-			if (ort->rt_mtu == nrt->rt_mtu)
+			/* If expiry has not been renewed by RA, and MTU is unchanged, skip */
+			if (ort->rt_mtu != nrt->rt_mtu ||
+				ort->rt_updated.tv_sec != nrt->rt_updated.tv_sec)
```
rsmarples, line 547: OK, this is not going to fly. We go out of our way to avoid changing the kernel routes when we can as some OS's will flush cached ARP entries and this is bad for network boot hosts where the whole rootfs is net mounted. When dhcpcd boots the updated time of a kernel route entry will be zero.

I think we should be comparing rt_expires and if there is too much deviation then we update the route. We can't use zero deviation as there cycles spent reading the kernel routes, building our own idea of of routes and stuff from acquisition.
If that doesn't work as expected then I think we will need to guard this code in __linux__ or HAVE_ROUTE_LIFETIME

Thoughts?

src/route.c
```diff
@@ -295,6 +295,8 @@ rt_desc(const char *cmd, const struct rt *rt)
 		    ifname, cmd,
 		    rt->rt_flags & RTF_REJECT ? " reject" : "",
 		    dest, prefix, gateway);
+	logdebugx("route flags %d lifetime %d mtu %d",
+		rt->rt_flags, rt->rt_expires, rt->rt_mtu);
```
rsmarples, line 299: This debug should be removed

rsmarples (changes requested):

src/route.c
ColinMcInnes: I agree, I'm ok with it only updating if it's outside a range. What do you think about 30s or more deviation? The minimum route lifetime is something like 200s.

I'm also ok gating it behind "HAVE_ROUTE_LIFETIME".

ColinMcInnes: I'll have to think about comparing rt_expires, but we can pull this into a "compare the route lifetime" function so the logic is a bit easier to read.

rsmarples:
@ColinMcInnes I've pushed some changes.

I finally got around to testing this on a Fedora box and spotted some issues which are now resolved. Please review my changes and comment if you agree or not. This cosmetic change is proving a lot of trouble and has revealed a very poor implementation which I've noted in comments.

src/route.c
```diff
@@ -502,18 +502,26 @@ rt_recvrt(int cmd, const struct rt *rt, pid_t pid)
...
 
-/* Compare route details */
+/* Compare miscellaneous route details */
 static bool
-rt_changed(struct rt *nrt, struct rt *ort)
+rt_cmp_misc(struct rt *nrt, struct rt *ort)
 {
 	/* MTU changed */
 	if (ort->rt_mtu != nrt->rt_mtu)
-		return true;
```
ColinMcInnes: Wouldn't this cause it to skip changes to the MTU?

rsmarples: @ColinMcInnes good spot! I wanted the function signature to be the same as rt_cmp for consistency so I changed the logic.

Should be ok now? Let me know!

ColinMcInnes: Looks good to me!

PR #442, opened 2025-01-28
Increase max IPv4 clientid.

gnaaman-dn:
Remove arbitrary limit, raising to the maximum representable by uint8

As discussed in https://github.com/NetworkConfiguration/dhcpcd/discussions/435

### DIFF
--- a/src/if-options.h
+++ b/src/if-options.h
@@ -57,7 +57,9 @@
 #define HOSTNAME_MAX_LEN	250	/* 255 - 3 (FQDN) - 2 (DNS enc) */
 #endif
 #define VENDORCLASSID_MAX_LEN	255
-#define CLIENTID_MAX_LEN	48
+/* Can't use the entire 255B allowed by the length field, since
+ * the first byte is the type of clientid */
+#define CLIENTID_MAX_LEN	254
 #define USERCLASS_MAX_LEN	255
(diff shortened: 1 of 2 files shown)


src/if-options.h
```diff
@@ -57,7 +57,7 @@
 #define HOSTNAME_MAX_LEN	250	/* 255 - 3 (FQDN) - 2 (DNS enc) */
 #endif
 #define VENDORCLASSID_MAX_LEN	255
-#define CLIENTID_MAX_LEN	48
+#define CLIENTID_MAX_LEN	255
```
rsmarples, line 60: I think we need to limit to 254 as the first byte of the clientid is the type.

gnaaman-dn: Whoopse, missed that.
Thank you

rsmarples: And this was a bad review, sorry.
The increase to 255 is fine because parse_hwaddr already handles setting the type to zero and adjusting the length as it goes. I've added comments for this and we now just define a DHCP_OPTION_MAX_LEN.

If this looks OK to you I will merge.

gnaaman-dn: Looks great to me, but I'll ask for a clarification:
All of the options you've changed are now 1-byte smaller.
Was this byte a NUL-terminator that is in practice unnecessary?

PR #446, opened 2025-01-29
Always send req for InfoRefreshTime option on Inform-Req

### LINKED ISSUE #445: INFORM6 rebind does not request lifetime option
On a regular inform-request lifetime renewal, it requests option 32 Inform-Req refresh lifetime and applies the lifetime in the reply.

Image

On a rebind, which triggers an inform-request, it does not request option 32, which results in it defaulting to  IRT_DEFAULT if the server does not send it.

Image

https://datatracker.ietf.org/doc/html/rfc8415#section-21.23

```
A DHCP client MUST request this option in the Option Request option
   (see [Section 21.7](https://datatracker.ietf.org/doc/html/rfc8415#section-21.7)) when sending Information-request messages.
```

Other options are requested normally. I will look into see if I can figure out what's different in the crafting of the inform-req message. All other options seem to be populated ok.

ColinMcInnes:
RFC8415 says client must always request Information Request refresh time option 32.
Client has other logic for handling if server does not include this in the reply.

Resolves #445

### DIFF
--- a/src/dhcp6.c
+++ b/src/dhcp6.c
@@ -1738,16 +1738,22 @@ static void
 dhcp6_startdiscover(void *arg)
 {
 	struct interface *ifp;
+	struct if_options *ifo;
 	struct dhcp6_state *state;
 	int llevel;
 	struct ipv6_addr *ia;
 
 	ifp = arg;
 	state = D6_STATE(ifp);
+	ifo = ifp->options;
 #ifndef SMALL
 	if (state->reason == NULL || strcmp(state->reason, "TIMEOUT6") != 0)
 		dhcp6_delete_delegates(ifp);
 #endif
+	/* Ensure we never request INFO_REFRESH_TIME,
+ 	 * this only belongs in Information-Request messages */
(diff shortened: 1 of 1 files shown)


src/dhcp6.c
```diff
@@ -4055,10 +4060,9 @@ dhcp6_start1(void *arg)
 		del_option_mask(ifo->requestmask6, D6_OPTION_RAPID_COMMIT);
 #endif
 
-	if (state->state == DH6S_INFORM) {
-		add_option_mask(ifo->requestmask6, D6_OPTION_INFO_REFRESH_TIME);
+	if (state->state == DH6S_INFORM)
 		dhcp6_startinform(ifp);
-	} else {
+	else {
 		del_option_mask(ifo->requestmask6, D6_OPTION_INFO_REFRESH_TIME);
```
rsmarples, line 4066: As we add the option in `dhcp6_startinform()`, we should delete it in `dhcp6_startdiscover()` so we handle all bases if we change from discover -> inform -> discover.

rsmarples (changes requested):
Small comment otherwise looks good.

PR #451, opened 2025-02-02
Uncomment 'hostname' in dhcpcd.conf

perkelix:
dhcpcd routinely fails to obtain an IP with a timeout if the DHCP server doesn't receive the "hostname" option. Let's uncomment it.

### DIFF
--- a/src/dhcpcd.conf
+++ b/src/dhcpcd.conf
@@ -5,7 +5,7 @@
 #controlgroup wheel
 
 # Inform the DHCP server of our hostname for DDNS.
-#hostname
+hostname
 
 # Use the hardware address of the interface for the Client ID.
 #clientid


ColinMcInnes:
What software is on the server? Is this a bug with the server software?

perkelix:
It might be server-specific.

ColinMcInnes:
> It might be server-specific.

Could also be config specific. It might be expecting a specific hostname to hand out an IP.

perkelix:
This only sets IPv4 and IPv6, no specific hostname:

https://salsa.debian.org/debian/dhcpcd/-/blob/debian/sid/debian/tests/timesyncd-ntp-servers-from-dhcp?ref_type=heads

ColinMcInnes:
> This only sets IPv4 and IPv6, no specific hostname:
> 
> https://salsa.debian.org/debian/dhcpcd/-/blob/debian/sid/debian/tests/timesyncd-ntp-servers-from-dhcp?ref_type=heads

Unless I've missed something, that looks like a dhcpcd config.

What does the dhcpd config look like?

perkelix:
Look carefully. This sets up dnsmasq to be able to test that dhcpcd correctly receives IP addresses and NTP server info.

perkelix:
Here's an example of such a timeout:
https://ci.debian.net/packages/d/dhcpcd/testing/amd64/57448052/

ColinMcInnes:
> Look carefully. This sets up dnsmasq to be able to test that dhcpcd correctly receives IP addresses and NTP server info.

A response is not the server config though. We know what is being sent, and what is coming back, but we do not know what the server is configured to send back.

What is the server, what version, and what config is it using? That might be important.

There is nothing in the DHCP RFCs that indicated the hostname should be required, so I'm unclear why the dhcp server is caring.

perkelix:
The server config is generated within this script. Everything else uses dnsmasq defaults. The test randomly fails with a timeout at the exact spot shown there, every now and then. Debian's release team considers this a CI issues since the test produces inconsistent results with the exact same config. The issue was first noticed at Ubuntu, which has a lot stricted CI testing.

ColinMcInnes:
> The server config is generated within this script.

What is the server, what version?

We're working with a black box here.

> Debian's release team considers this a CI issues since the test produces inconsistent results with the exact same config.

I agree it's a CI issue, I'm just not clear which end of the system is at fault here. Is dhcpcd causing the server to hiccup, or is the server hiccuping when dhcpcd does not ask for hostname?

According to the RFCs, dhcpcd should be right, hostname has nothing to do with a lease. So what is the disconnect here? What is the server expecting that dhcpcd is breaking within the rules of DHCP?

What version of dhcp server is Debian testing against, with what dhcp server config?

perkelix:
Both dhcpcd and dnsmasq versions keep on evolving since CI takes place against Testing and Unstable releases.

I've asked the dnsmasq maintainers (one of which happens to be upstream) for advice. I have yet to get an answer.

ColinMcInnes:
> Both dhcpcd and dnsmasq versions keep on evolving since CI takes place against Testing and Unstable releases.
> 
> I've asked the dnsmasq maintainers (one of which happens to be upstream) for advice. I have yet to get an answer.

dnsmasq is a client service... what is the dhcp server the CI test is using?

perkelix:
DHCP service is provided by the dnsmasq configured in that CI testing script.

ColinMcInnes:
I guess a better question might be what is different between dhcpcd 10.1.0 and Debian dhcpcd 1:10.1.0-4?

DanielG:
On Sun, Feb 02, 2025 at 11:23:45PM -0800, Colin McInnes wrote:
> dnsmasq is a client service... what is the dhcp server the CI test is using?

Colin, dnsmasq is a widely deployed fully integrated DHCP server and DNS
resolver (among other things). Popular for example on embedded devices
(CPEs) and in hypervisors such as libvirt, lxc, incus. See also `apt-cache
rdepends dnsmasq-base` and https://thekelleys.org.uk/dnsmasq/doc.html for
more reading material.

You may be thinking of the NetworkManager client integration with dnsmasq
where, indeed, it's used only for it's resolver function. That's not all it
can do :-)

--Daniel

perkelix:
> @perkelix Have you heard anything back from dnsmasq maintainers? Is this something they could recommend be added in a dhcpcd.conf specific to their CI test, since they're asking for something that is not requried by the DHCP specification?

I haven't heard back. Anyhow, as Debian seems adamant on this being uncommented, I build the package with a simple patch that removes the hash in front of the option. It's a minimal enough change to the stock config that I can live with that.

PR #486, opened 2025-03-13
dhcpcd.conf: Switch clientid/duid default.

### LINKED ISSUE #485: Consider changing default dhcpcd.conf duid vs. clientid behavior.
It seems known that this breaks some clients (2c8889dcf60359ccc00304052ec19f4a74aada2d).

Over the last couple months I've found at least 3 (I decided to revert just to see if it was a one-off):

 * Helsinki Airport
 * SFO Airport
 * Apple public wifi

I don't travel _all_ that much, and I've probably missed others. It might be worth just changing the default?

emilio:
Not sure if y'all want to do this, but if you do here is the patch :)

Closes #485

### DIFF
--- a/src/dhcpcd.conf
+++ b/src/dhcpcd.conf
@@ -8,12 +8,12 @@
 #hostname
 
 # Use the hardware address of the interface for the Client ID.
-#clientid
+clientid
 # or
 # Use the same DUID + IAID as set in DHCPv6 for DHCPv4 ClientID as per RFC4361.
 # Some non-RFC compliant DHCP servers do not reply with this set.
-# In this case, comment out duid and enable clientid above.
-duid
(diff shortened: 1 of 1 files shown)


perkelix:
Doing this only makes sense for hosts that are not dual-stack. On dual-stack hosts, duid is exactly what we want.

perkelix:
The key problem is that IPv6 (especially DHCPv6) is essentially an afterthought. As such, many routers only implement support for early drafts of the specification that presupposes EUI64 addresses. Additionally, many routers only support an antiquated version of DHCPv4 that is not aware of duid.

emilio:
Ok closing as since per #485 this is somewhat on the radar. Thanks!

PR #520, opened 2025-07-24
FreeBSD: Introduce `vlan_pcp` option for VLAN PCP tagging

rcmcdonald91:
This is needed to achieve feature parity with dhclient(8) in FreeBSD.

https://reviews.freebsd.org/D31263

```
This allows the use of VLAN PCP in dhclient, which is required for
certain ISPs (such as Orange.fr).
```

Still needs documentation.

### DIFF
--- a/src/bpf.c
+++ b/src/bpf.c
@@ -162,6 +162,9 @@ bpf_open(const struct interface *ifp,
 #ifdef BIOCIMMEDIATE
 	unsigned int flags;
 #endif
+#ifdef BIOCSETVLANPCP
+	unsigned int vlan_pcp;
+#endif
 #ifndef O_CLOEXEC
 	int fd_opts;
 #endif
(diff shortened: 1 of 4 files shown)


src/bpf.c
```diff
@@ -162,6 +162,9 @@ bpf_open(const struct interface *ifp,
 #ifdef BIOCIMMEDIATE
 	unsigned int flags;
 #endif
+#ifdef BIOCSETVLANPCP
+	u_int vlan_pcp;
```
rsmarples, line 166: Could we use the same style and use unsigned int please?

src/bpf.c
```diff
@@ -207,6 +210,14 @@ bpf_open(const struct interface *ifp,
 	if (ioctl(bpf->bpf_fd, BIOCSETIF, &ifr) == -1)
 		goto eexit;
 
+#ifdef BIOCSETVLANPCP
+	if (ifp->options->vlan_pcp > -1) {
```
rsmarples, line 214: A better test would be != -1 as it's less CPU work.

rsmarples (changes requested):
Could you add an entry to dhcpcd.conf(5) as well please?

PR #525, opened 2025-09-06
add uninstall target to makefile

equwal:
uninstall


perkelix:
I would be very tempted to revert the commit. @rsmarples, what do you think?

PR #533, opened 2025-09-27
privsep: enforce message boundaries with MSG_EOR on our messages

### LINKED ISSUE #530: dhcpcd fails to get address on FreeBSD STABL-15, fails to exit
Hello.

Just upgraded to STABLE-15 to test stuff. Custom kernel, customized base, but dhcpcd was working flawlessly on STABLE-14 just before upgrade.

compatXX ports are installed, system is cleaned after upgrade (yes | make delete-old; yes | make delete-old-libs)

dhcpcd-10.2.4 starting
spawned privileged proxy on PID 4291
spawned controller proxy on PID 4293
DUID 00:01:00:01:2e:bc:cf:ae:fa:89:3b:22:6b:2a
sandbox: capsicum
lo0: ignoring due to interface type and no config
re0: executing: /usr/local/libexec/dhcpcd-run-hooks PREINIT
re0: executing: /usr/local/libexec/dhcpcd-run-hooks NOCARRIER
wlan0: executing: /usr/local/libexec/dhcpcd-run-hooks PREINIT
wlan0: executing: /usr/local/libexec/dhcpcd-run-hooks NOCARRIER
no interfaces have a carrier
re0: waiting for carrier
wlan0: carrier acquired
wlan0: interface updated
wlan0: connected to Access Point: guru
wlan0: executing: /usr/local/libexec/dhcpcd-run-hooks CARRIER
wlan0: IAID 4b:6c:f5:61
wlan0: adding address fe80::a3c6:fe0:2c0a:6e6a
wlan0: delaying IPv6 Router Solicitation for LL address
wlan0: delaying IPv4 for 0.2 seconds
wlan0: reading lease: /var/db/dhcpcd/wlan0-guru.lease
wlan0: discarding expired lease
wlan0: soliciting a DHCP lease
wlan0: spawned BPF BOOTP on PID 4430
wlan0: delaying IPv6 router solicitation for 0.2 seconds
wlan0: soliciting an IPv6 router
wlan0: sending DISCOVER (xid 0x6c66c830), next in 8.5 seconds
wlan0: process BPF BOOTP already started on pid 4430
wlan0: probing for an IPv4LL address
wlan0: probing for 169.254.46.199
wlan0: spawned BPF ARP 169.254.46.199 on PID 4431
wlan0: sending Router Solicitation
wlan0: ARP probing 169.254.46.199 (2 of 3), next in 1.7 seconds
wlan0: ARP probing 169.254.46.199 (3 of 3), next in 2.0 seconds
wlan0: sending Router Solicitation
wlan0: using IPv4LL address 169.254.46.199
wlan0: ARP announcing 169.254.46.199 (1 of 2), next in 2.0 seconds
wlan0: adding route to 169.254.0.0/16
wlan0: adding default route
wlan0: executing: /usr/local/libexec/dhcpcd-run-hooks IPV4LL
wlan0: ARP announcing 169.254.46.199 (2 of 2)
wlan0: sending DISCOVER (xid 0x6c66c830), next in 15.7 seconds
wlan0: process BPF BOOTP already started on pid 4430
wlan0: sending Router Solicitation
wlan0: sending DISCOVER (xid 0x6c66c830), next in 31.4 seconds
wlan0: process BPF BOOTP already started on pid 4430
wlan0: sending DISCOVER (xid 0x6c66c830), next in 64.1 seconds
wlan0: process BPF BOOTP already started on pid 4430

When issuing dhcpcd -x this happens:

sending signal TERM to pid 4289
waiting for pid 4289 to exit
pid 4289 failed to exit

Base dhclient is getting IP in a second or two. Same applies to other interfaces, like ue0.

Big thanks for your project!

glebius:
The nature of the SOCK_SEQPACKET, that privsep modules uses, is stream. See:

https://pubs.opengroup.org/onlinepubs/9799919799/functions/V2_chap02.html#tag_16_10_06

To guarantee that a reader will never read two messages in one read operation, the writer shall put end of record markers.

The problem exposed itself in FreeBSD 15.0 that started to follow the specification better than before.

Other SOCK_SEQPACKET usage considerations: a) as long as our reader provides a receive buffer that would fit the largest message our writer would ever send, we are good with regards to not a reading a partial message b) as long as our writer always write full messages with one write, we don't need use of MSG_WAITALL in reader.

Fixes #530

### DIFF
--- a/src/privsep.c
+++ b/src/privsep.c
@@ -934,7 +934,9 @@ ps_sendpsmmsg(struct dhcpcd_ctx *ctx, int fd,
 	} else
 		iovlen = 1;
 
-	len = writev(fd, iov, iovlen);
+	len = sendmsg(fd,
+	    &(struct msghdr){ .msg_iov = iov, .msg_iovlen = iovlen }, MSG_EOR);
+
 	if (len == -1) {
 		if (ctx->options & DHCPCD_FORKED &&
 		    !(ctx->options & DHCPCD_PRIVSEPROOT))
@@ -1063,7 +1065,9 @@ ps_sendcmdmsg(int fd, uint16_t cmd, const struct msghdr *msg)
 	    psm.ps_namelen + psm.ps_controllen + psm.ps_datalen + cmsg_padlen;
 	if (psm.ps_datalen != 0)
 		memcpy(p, msg->msg_iov[0].iov_base, psm.ps_datalen);
-	return writev(fd, iov, __arraycount(iov));
+	return sendmsg(fd,
+	    &(struct msghdr){ .msg_iov = iov, .msg_iovlen = __arraycount(iov) },
+	    MSG_EOR);
 
 nobufs:
 	errno = ENOBUFS;


kworr:
Process can't start:

```
main: control_open: Connection refused
dhcpcd-10.2.4 starting
spawned privileged proxy on PID 22026
spawned network proxy on PID 22027
spawned controller proxy on PID 22028
/var/db/dhcpcd/duid: Operation not supported
sandbox: capsicum
ps_root_getifaddrs: Operation not supported
main: control_stop: Operation not supported
network proxy exited from PID 22027
control proxy exited from PID 22028
dhcpcd exited
```

glebius:
I just retried dhcpcd master (with my patch) on FreeBSD CURRENT, I can't reproduce the problem you got.

glebius:
Can you please share ktrace dump of your failed start with me at glebius@FreeBSD.org?

perkelix:
@glebius thanks for looking into this FreeBSD-specific issue. If you can spot any other pending bug specifically affecting FreeBSD, you're welcome to chip in. Btw, are you the one who subscribed CodeRabbit to this bug?

kworr:
Disregard breakage, that was my dirty hands. Patch works.

rsmarples:
> @perkelix Precisely speaking, this is not really a FreeBSD-specific one. It is a historical bug that FreeBSD SOCK_SEQPACKET would create implicit message boundaries on writes. I heard Linux also has this bug and there are applications that rely on it. This behavior truly contradicts the definition of the SOCK_SEQPACKET and makes it almost the same as SOCK_DGRAM. To make things worse, the Single UNIX Specification while very clearly describes SOCK_SEQPACKET in the main page on sockets, has documentation bugs in recv() functions. This will be corrected in the next version of SUS.

I just read all the man pages around this and I have to ask are you sure? Because there is no clear indication anywhere that you *must* use MSG_EOR.
Here's what FreeBSD-16 socket(2) says:
> SOCK_SEQPACKET  sockets	employ	the  same  system calls	as SOCK_STREAM
       sockets.	 The only difference is	that [read(2)](https://man.freebsd.org/cgi/man.cgi?query=read&sektion=2&apropos=0&manpath=FreeBSD+16.0-CURRENT) calls  will  return  only
       the  amount of data requested, and any remaining	in the arriving	packet
       will be discarded.

So if we can use read, then we can use write which implies implicit message boundaries.

If this is a problem, then we could more easily change from SOCK_SEQPACKET to SOCK_STREAM and not bother.
We send a fixed header for each message with enough blurb about how much trailing data we have, so we implicitly know our own message boundaries.

glebius:
I'm working on the socket(2) manual page right now. It has quite a lot of outdated information and documentation of SOCK_SEQPACKET is just the worst.

There is no must in use of MSG_EOR in general. But if you want message boundaries, you need it.

You could change from SOCK_SEQPACKET to SOCK_DGRAM, cause your use is much closer to datagram, than to stream. Your expectation is that read size == write size, which is datagram thing, not stream. But you'd better stay with SOCK_SEQPACKET, it is superior. Just use it correctly.

rsmarples:
OK, so can you explain why I need this? Is it because somehow we are tripping on a boundary we shouldn't?

glebius:
I'm not sure your refactoring is correct. With my original change I browsed around the code and made sure that the sendmsg(MSG_EOR) are done only on the file descriptors that are SOCK_SEQPACKET. Putting this flag on other kinds of descriptors doesn't make sense. FreeBSD most likely will just ignore it, so it is not a critical mistake. Not sure about other OSes. I'd better use sendmsg(MSG_EOR) only on those code paths that work with SOCK_SEQPACKET only.

src/logerr.c
```diff
@@ -215,18 +217,23 @@ vlogmessage(int pri, const char *fmt, va_list args)
...
-
-		memcpy(buf, &pri, sizeof(pri));
-		pid = getpid();
-		memcpy(buf + sizeof(pri), &pid, sizeof(pid));
-		len = vsnprintf(buf + sizeof(pri) + sizeof(pid),
-		    sizeof(buf) - sizeof(pri) - sizeof(pid),
-		    fmt, args);
-		if (len != -1)
-			len = (int)write(ctx->log_fd, buf,
-			    ((size_t)++len) + sizeof(pri) + sizeof(pid));
```
glebius, line 229: There is no need to rewrite everything into iovec form. You could use just:
```
send(ctx->log_fd, buf, ((size_t)++len) + sizeof(pri) + sizeof(pid), MSG_EOR);
```

glebius:
Yep, looks like the logging one also needs MSG_EOR, otherwise logged messages may glue together.

src/logerr.c
```diff
@@ -413,8 +413,13 @@ logreadfd(int fd)
 	};
 
 	len = (int)recvmsg(fd, &msg, 0);
```
glebius, line 415: You might want to add MSG_WAITALL flag here. That will guarantee that read operation will wait either provided buffer is filled or a message boundary is observed. That will guarantee against truncated messages.

P.S. With the existing code I doubt truncation can happen, as you always write log messages in a single write operation.
P.P.S. I don't know how if on Linux MSG_WAITALL will stop reads on message boundaries.

rsmarples: That's fair. I'll be testing on other OS's before this gets merged as it touches critical paths.

src/logerr.c
```diff
@@ -390,24 +399,30 @@ int
...
+		.msg_iovlen = sizeof(iov) / sizeof(iov[0])
+	};
+
+	len = (int)recvmsg(fd, &msg, MSG_WAITALL);
+	if (len == -1 || len == 0)
 		return -1;
-
-	/* Ensure we have pri, pid and a terminator */
-	if (len < (int)(sizeof(pri) + sizeof(pid_t) + 1) ||
-	    buf[len - 1] != '\0')
```
glebius, line 402: Now we are no longer checking that the other side sent NUL terminated message. I don't know if this is intentional change. But it is unrelated to the gist of this changeset.

rsmarples:
@glebius @kworr @perkelix OK, I think we're good here now. All the sockets of type SOCK_SEQPACKET are now always sent with MSG_EOR, code works and compiles with no warnings on all platforms.

rcmcdonald91:
Glad to see traction here! Thanks everyone

PR #556, opened 2025-11-17
privsep: Ensure we recv for real after a successful recv MSG_PEEK

### LINKED ISSUE #555: SIOCSPFXFLUSH_IN6: No buffer space available after updating to 10.3.0 (FreeBSD 14.3)
After years of trouble-free operation of the following dhcpcd setup, it fails with "**SIOCSPFXFLUSH_IN6: No buffer space available**"
This problem arose after upgrading to dhcpcd 10.3.0 on my FreeBSD 14.3 installation.

```
quiet
ipv6only
noipv6rs
release
background
vendorclassid
slaac private

clientid
allowinterfaces ng0
nohook test,hostname,resolv.conf,ntp.conf

option interface_mtu
option classless_static_routes
require dhcp_server_identifier

interface ng0
  ipv6rs
  ia_pd 1/::/56 bridge0/0/64
```

rsmarples:
Adjust the code flow so that the same errors would be caught after the final recv.
This ensures we read what is really meant for us and not something silly.

Should fix #555.

### DIFF
--- a/src/privsep-root.c
+++ b/src/privsep-root.c
@@ -71,20 +71,18 @@ struct psr_ctx {
 	struct psr_error psr_error;
 	size_t psr_datalen;
 	void *psr_data;
-	size_t psr_mdatalen;
-	void *psr_mdata;
-	bool psr_usemdata;
+	bool psr_mallocdata;
 };
 
 static ssize_t
-ps_root_readerrorcb(struct psr_ctx *psr_ctx)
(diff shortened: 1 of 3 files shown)


rsmarples:
> Does this one seem ready to merge too?

I would like some feedback on it first as I still don't know if it fixes people issues as I cannot replicate the problem.

PR #581, opened 2026-01-09
common: Cast via uintptr_t rather than unsigned long in UNCONST

jrtc27:
On ILP32 and LP64 systems, unsigned long is sufficient to round-trip pointers, but on LLP64 systems it is too small, and on systems like CHERI where pointers are capabilities, it does not propagate capability metadata, and casting back to a pointer to yield a pointer that cannot be dereferenced.

Instead, cast via uintptr_t to ensure we always use an integral type that can losslessly round-trip pointers.

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -55,7 +55,7 @@
 #define MAX(a,b)		((/*CONSTCOND*/(a)>(b))?(a):(b))
 #endif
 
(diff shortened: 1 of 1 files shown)


rsmarples:
Thanks!

PR #592, opened 2026-03-27
Change eghooks to be opt-in

### LINKED ISSUE #575: Allow disabling eghooks completely
In its current form (10.3.0), dhcpcd doesn't allow skipping EGHOOKS at all. I tried `--without-eghooks="wpa_supplicant.conf timezone.conf lookup-hostname.conf"` and `--with-eghooks=""` but the 3 sample scripts get shipped regardless. Given how the 3 examples are unlikely to be needed on recent distributions, it should be possible to not ship them at all. This would eliminate the ambiguity between the two hook directories, since users have previously reported not understanding why there seemingly is two hook directories, one in libexec, one in share.

ColinMcInnes:
If not using --with-eghooks, do not install any example hooks in DATADIR

Resolves #575

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -176,13 +176,15 @@ Example: `make DESTDIR=/usr/src/contrib/dhcpcd import-src`
 
 ## Hooks
 Not all the hooks in dhcpcd-hooks are installed by default.
-By default we install `01-test`, `20-resolv.conf`and `30-hostname`.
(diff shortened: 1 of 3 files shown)


perkelix:
@rsmarples can we merge this?

perkelix:
@rsmarples Can we merge this?

perkelix:
@rsmarples can we also return to this one?

Issue #574, opened 2025-12-29
--with-hooks="timesyncd.conf" should uncomment 'option ntp_servers'


perkelix:
As [reported](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1123962) at Debian, dhcpcd apparently started commenting out 'option ntp_servers' at commit aa1a876. This results in dhcpcd shipping without NTP support even in cases when `--with-hooks="timesyncd.conf"` or `--with-hooks="ntp.conf"` was passed to `./configure`. IMHO, the build should uncomment the option in `dhcpcd.conf` if either hook was explicitly configured.

AdamPerkins:
Ignoring the previous AI nonsense I strongly recommend 'option ntp_servers' is active by default.

I don't understand why it was commented out without any explanation in the commit.

ColinMcInnes:
I think configure/Makefile automatically uncomment that option via the Makefile if --enable-ntp is used.

Is that the case?

perkelix:
It seems to do so. Thanks.

However, IMHO, it would make more sense for any of... 

- `--with-hooks="timesyncd.conf"`
- `--with-hooks="ntp.conf"`
- the NTP daemon detection traps in `./configure`

...to uncomment it for us, since including the hooks but shipping the config with `ntp_servers` commented out is counter-productive.

ColinMcInnes:
To be fair, configure doesn't know that the hook filename is reflective of the need for ntp functionality. However, I'm open to it being a little more liberal on detection since those two conf files are explicitly included in the distro, so we know they're linked to ntp.

Another alternative could be to add "--with-timesyncd" and "--with-ntpd" that would explicitly enable ntp, the ntp option, and add the appropriate hook. What do you think about that @perkelix ?

- --with-timesyncd (enables ntp and chooses the timesyncd.conf hook)
- --with-ntpd (enables ntp and chooses the ntp.conf hook)
- --with-ntp (enables ntp only, maintainer has to specify a hook with --with-hook)
- the NTP daemon detection traps in ./configure (chooses the hook based on which daemon it detected)

perkelix:
This sounds like a plan. The daemon detection traps would also need to enable ntp.

For `ntpd`, we might need to update our scripts to test for different locations for ntpd.conf, since e.g. Debian puts these into their own subdirectories inside `/etc`.

ColinMcInnes:
[...]

PR #597, opened 2026-04-06
Enable ntp if autodetect ntpd or timesyncd

ColinMcInnes:
If --enable-ntpd or --enable-timesyncd, assume --enable-ntp and
choose the appropriate hook file.

Will complain if --enable-ntpd or --enable-timesyncd are both set,
allows override of autodetect.

Autodetect will enable ntp functionality unless explicitly told not
to via --disable-ntp. If --disable-ntp is used, no time hooks
are installed.

Addresses #574

### DIFF
--- a/configure
+++ b/configure
@@ -40,7 +40,6 @@ SANITIZE=no
 STATUSARG=
 OPENSSL=
 NTPD=
-CHRONYD=
 TIMESYNCD=
 
 DHCPCD_DEFS=dhcpcd-definitions.conf
@@ -1701,13 +1700,14 @@ if [ -z "$STATUSARG" ]; then
 fi
 echo "STATUSARG=	$STATUSARG" >>config.mk
 
+if [ "$NTPD" = "yes" ] && [ "$TIMESYNCD" = "yes" ]; then
+	echo "Warning: --enable-ntpd and --enable-timesyncd are both set" >&2
+	exit 1
+fi
+
 HOOKS=
 if ! $HOOKSET; then
     if ! [ "$NTP" = "no" ]; then
-	    if [ "$NTPD" = "yes" ] && [ "$TIMESYNCD" = "yes" ]; then
-			echo "Warning: --enable-ntpd and --enable-timesyncd are both set" >&2
-			exit 1
-		fi
 		printf "Checking for ntpd ... "
 		NTPD_PATH=$(_which ntpd)
 		if [ -n "$NTPD_PATH" ]; then
@@ -1746,6 +1746,11 @@ if ! $HOOKSET; then
 			echo "$TIMESYNCD_PATH (Selecting 50-timesyncd.conf)"
 		else
 			echo "not found"
+		fi
+
+		 # Warn if both ntpd/chronyd and timesyncd are detected
+		if [ "$NTPD" = "yes" ] && [ -n "$TIMESYNCD_PATH" ]; then
+			echo "Both ntpd/chronyd and timesyncd detected; NTP will default to ntpd"
 		fi
 
 		# Choose hook based on enabled time service
--- a/configure
+++ b/configure
@@ -1703,22 +1703,9 @@ if [ -z "$STATUSARG" ]; then
 fi
 echo "STATUSARG=	$STATUSARG" >>config.mk
 
-if [ "$NTPD" = "yes" ] && [ "$TIMESYNCD" = "yes" ]; then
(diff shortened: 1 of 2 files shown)


configure
```diff
@@ -1790,9 +1816,24 @@ if ! $HOOKSET; then
...
-	# --enable-ntp
+
+if [ -z "$NTP" ]; then
+	if [ "$NTPD" = "yes" ] || [ "$TIMESYNCD" = "yes" ]; then
+		NTP=yes
+	fi
+fi
+if [ "$NTP" = "yes" ]; then 
+	# --enable-ntp or ntp/timesyncd service was autodetected
+	printf "Enabling NTP support for ... "
```
ColinMcInnes: What if enable-ntp but not either of these (user has their own hook defined). Make this line more clear.

ColinMcInnes: done

perkelix:
Why would it complain if both --enable-ntpd and --enable-timesyncd are set? A distro might ship both hooks.

ColinMcInnes:
> Why would it complain if both --enable-ntpd and --enable-timesyncd are set? A distro might ship both hooks.

I was focusing on fixing the autodetect, but that's a good point, you might not know at compile time which one you'd end up with at the end point, so if --enable is used it should not complain and allow both.

@perkelix if it's configured for both, I see that timesyncd.conf will exit if it's not present at runtime, but I don't see ntpd.conf doing the same, is that ok? that might result in it complaining loudly if timesyncd is installed on the target system.

```
./configure --enable-ntpd
configure args: --enable-ntpd
Deriving operating system from ... x86_64-unknown-linux
...
Testing for dlopen ... yes
Checking for ntpd ... selected via --enable-ntpd
Checking for timesyncd ... /usr/lib/systemd/systemd-timesyncd (Selecting 50-timesyncd.conf)
Checking for ypbind ... not found
Assuming ypbind is Linux
Enabling NTP support

   SYSCONFDIR =		/etc
   SBINDIR =		/sbin
   LIBDIR =		/lib
   LIBEXECDIR =		/libexec
   DBDIR =		/var/db/dhcpcd
   RUNDIR =		/var/run/dhcpcd
   MANDIR =		/usr/share/man
   DATADIR =		/usr/share
   HOOKSCRIPTS =	50-ntp.conf 50-timesyncd.conf
   EGHOOKSCRIPTS =	50-yp.conf
   STATUSARG =		
   PRIVSEPUSER =	dhcpcd
```

perkelix:
Obviously ntp.conf should also exit if it doesn't find the right daemon on the system.

ColinMcInnes:
> Obviously ntp.conf should also exit if it doesn't find the right daemon on the system.

It looks like all it really does is configure/deconfigure the ntp.conf file for an ntp daemon and quietly requests the system restart the service, throwing errors to dev/null. So I think we're ok there.

configure
```diff
@@ -1698,41 +1705,68 @@ echo "STATUSARG=	$STATUSARG" >>config.mk
...
+		printf "Checking for timesyncd ... "
+		if [ -z "$TIMESYNCD" ]; then
+			TIMESYNCD_PATH=
+			for x in /usr/lib/systemd/systemd-timesyncd; do
+				if [ -x "$x" ]; then
+					TIMESYNCD=yes
+					TIMESYNCD_PATH=$x
+					NTP=yes
+					break
+				fi
```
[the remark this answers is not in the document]
ColinMcInnes: Should update the timesyncd.conf as well, have it check both paths.

perkelix:
@rsmarples any comment on the above?

perkelix:
@rsmarples Can we merge this?

perkelix:
@rsmarples can we return to this one?

configure
```diff
@@ -1829,41 +1709,68 @@ echo "STATUSARG=	$STATUSARG" >>config.mk
...
-		echo "not found"
-	fi
-	if [ -n "$NTPD" ] || [ -n "$CHRONYD" ]; then
-		HOOKS="$HOOKS${HOOKS:+ }50-ntp.conf"
-	fi
-	# Warn if both are detected
-	if [ -n "$NTPD" ] && [ -n "$CHRONYD" ]; then
-		echo "NTP will default to $NTPD"
-	fi
+    if [ "$NTP" = "yes" ]; then
```
rsmarples, line 1712: This line needs to go.
We still want to autodetect what ntp we have available and provide a hook script for it even if it's disabled by default.

PR #617, opened 2026-05-30
pidfile: Import new pidfile from NetBSD

rsmarples:
Adapt code, chroot and sandboxing is more readily supported.

ctx->pidfile is now malloced rather than a fixed size.


rsmarples:
@coderabbitai full review combat/pidfile.c

PR #641, opened 2026-06-17
script: Use correct buffer length variable

craciunoiuc:
Final call to the control queue used the wrong 'buflen' variable. Change so the function stays consistent.


rsmarples:
@coderabbitai review

rsmarples:
Thanks!

PR #651, opened 2026-06-20
ARP: check we have enough to read the frame header

rsmarples:
Reported by NVIDIA Project Vanessa


rsmarples:
@coderabbitai review

PR #666, opened 2026-06-22
IPv6: Fix numerous issues extending temporary address times

rsmarples:
Fix a potential integer underflow if pltime is less than aquisition time.
Fix potential time truncation to uint32_t.

Reported by NVIDIA Project Vanessa


rsmarples:
@coderabbitai review

PR #677, opened 2026-06-27
A few changes to the Hurd port based on hooks

jlledom:
This adds a few changes to the pfinet hook. I'll explain further in comments below.

### DIFF
--- a/configure
+++ b/configure
@@ -500,6 +500,7 @@ netbsd*)
 hurd*)
 	echo "CPPFLAGS+=	-D_GNU_SOURCE" >>$CONFIG_MK
 	echo "DHCPCD_SRCS+=	if-hurd.c" >>$CONFIG_MK
+	echo "LDADD+=		-lpthread" >>$CONFIG_MK
 	BPF_OS="bpf-hurd.c"
 	# No INET6 support
 	INET6=no
--- a/hooks/10-pfinet
+++ b/hooks/10-pfinet
@@ -1,15 +1,18 @@
 # SPDX-License-Identifier: BSD-2-Clause
 # Copyright (c) 2026 Roy Marples <roy@marples.name>
 
-# Configure the Hurd PFINET translator.
+# Configure the Hurd network translator (pfinet or lwip).
 # It always on socket 2.
 
-case "$(showtrans /servers/socket/2)" in
-/hurd/pfinet" "*) pfinet=true;;
-*) pfinet=false;;
+trans=$(showtrans /servers/socket/2 2>/dev/null)
+[ -z "$trans" ] && trans=$(fsysopts /servers/socket/2 2>/dev/null)
+
+case "$trans" in
+*/pfinet" "*|*/lwip" "*) hurd_net=true;;
+*) hurd_net=false;;
 esac
 
-if $pfinet && $if_configured && [ "$ifxname" != "lo" ] ; then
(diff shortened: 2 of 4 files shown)


hooks/10-pfinet
```diff
@@ -1,15 +1,18 @@
...
 
-case "$(showtrans /servers/socket/2)" in
-/hurd/pfinet" "*) pfinet=true;;
-*) pfinet=false;;
+trans=$(showtrans /servers/socket/2 2>/dev/null)
+[ -z "$trans" ] && trans=$(fsysopts /servers/socket/2 2>/dev/null)
+
+case "$trans" in
+*/pfinet" "*|*/lwip" "*) hurd_net=true;;
+*) hurd_net=false;;
```
jlledom, line 12: A few changes here:

1. `showtrans` only returns passive translators, that is, when the directory entry in the FS has a record to execute the translator when the first RPC is received. However, we can also have active translators running in a node that are not installed as passive translators. Those are detected by `fsysopts`.
2. Now it also detects `lwip`. I made some changes in the lwip translator so it mimics how pfinet responds to fsysopts. I'll send the patches to the hurd mailing list. But since now they are compatible, we can reuse the same hook from both.
3. The pattern has been modified a bit so translators are detected also out of their default path at `/hurd/*`.

hooks/10-pfinet
```diff
@@ -20,7 +23,7 @@ if $pfinet && $if_configured && [ "$ifxname" != "lo" ] ; then
 		fi
 		fsysopts /servers/socket/2 -i "$ifxname" \
 			-a "$new_ip_address" $flags
-	elif $if_down; then
+	elif $if_down || { [ "$reason" = STOPPED ] && ! $if_persistent; }; then
```
jlledom, line 26: In my previous patches, which configured lwip/pfinet via ioctls, having `persistent == false` and `release == false` made dhcpcd unconfigure the interface on exit. However, this doesn't work when using the hook. With the hook, `release == true` unconfigures the stack in the kernel on exit but `persistent == false` + `release == false` doesn't. IMO it makes more sense to unconfigure always when `persistent == false`, and it seems to mimic how BSD behaves. Or am I missing something?

src/script.c
```diff
@@ -358,6 +358,9 @@ make_env(struct dhcpcd_ctx *ctx, const struct interface *ifp,
 	if (efprintf(fp, "if_configured=%s",
 		ifo->options & DHCPCD_CONFIGURE ? "true" : "false") == -1)
 		goto eexit;
+	if (efprintf(fp, "if_persistent=%s",
+		ifo->options & DHCPCD_PERSISTENT ? "true" : "false") == -1)
+		goto eexit;
```
jlledom, line 363: Just a new var so the `pfinet` hook knows the value of the `persistent` option

configure
```diff
@@ -500,6 +500,7 @@ netbsd*)
 hurd*)
 	echo "CPPFLAGS+=	-D_GNU_SOURCE" >>$CONFIG_MK
 	echo "DHCPCD_SRCS+=	if-hurd.c" >>$CONFIG_MK
+	echo "LDADD+=		-lpthread" >>$CONFIG_MK
```
jlledom, line 503: For some reason this is not added automatically in my local environment. It could be just my box, if it complies for you in the Hurd without it, this is probably not needed.

perkelix:
Is this is sync with the current 10.5.x release?

PR #678, opened 2026-06-29
options: Introdce policy groups

rsmarples:
Remove the old bitmask array which we used to store which options to request, remove, require, etc and replace with a more generic policy group.

The policy group structure holds an array of option numbers in each list - request, remove, require, etc.
Add a suite of helper functions around this to allow adding, removing, checking, walking, etc.

This allows a much easier use of this in DHCP processing but more importantly allows us to remove or reject options we do not have a definition for.


rsmarples:
@coderabbitai review

PR #680, opened 2026-07-05
control: remove unprivileged socket

rsmarples:
Instead workout credentials from the connected socket by getpeereid(2).
Add `readgroup` to dhcpcd.conf which controls who can read interface data from the control socket.
Each control socket command will now return an errno after the command has been processed to indicate the result of the command - for example EPERM.
Remove the limit of only one connected control client.

This is a breaking change for socket consumers.


rsmarples:
@coderabbitai review

rsmarples:
@coderabbitai review

rsmarples:
@coderabbitat review

PR #686, opened 2026-08-02
keep acquisition timestamp when route isn't updated

### LINKED ISSUE #681: Route renewal broken again
#465 seems to be broken again in 10.3.2: I observe route expiration marching down to zero despite receiving RAs from the upstream router with route expirations that should reset the timer; then I see a `pid 0 deleted default route via...` message; followed a few seconds later by the route being added back.

(Note that I observe this behavior on two networks, Comcast and Spectrum, whose routers send unsolicited RAs at essentially the peak rate permitted by the neighbor discovery RFC 4861 (every three seconds). In each case, there appear to be two routers that both send at the maximum permitted rate, which means clients receiving, in aggregate, RAs more than once every three seconds. It is unclear based on my reading whether this is a misconfiguration or not; but regardless, it works this way on at least two major cable internet providers, which means the software probably just needs to deal.)

Freax13:
RFC 4861 specifies that router advertisements should be sent no more than once every MIN_DELAY_BETWEEN_RAS (3 seconds). In practice though, there are several ISPs (Comcast, Spectrum, Vodafone) that send out router advertisements at higher rates.

Previously this caused dhcpcd to never update the route until it eventually expired and was removed by the kernel.

This happened because the rt_acquired timestamp was out of sync with the kernel expiry tracking. rt_acquired was updated every time a router advertisement was processed **even when the route wasn't updated in the kernel**. When router advertisements were sent out frequently, rt_acquired was updated frequently, the difference between the old and new rt_acquired values never became great enough for rt_cmp_lifetime to return 1 and the kernel's routing table was never updated.

To fix this, keep the old rt_acquired value iff the kernel's routing table wasn't updated. This ensures that rt_acquired stays in sync with the expiry value in the kernel's routing table.

Closes #681
Cc @squarooticus

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,9 +51,6 @@ jobs:
     - name: Update packages
       run: sudo apt update
 
-    - name: Upgrade packages
-      run: sudo apt upgrade -y
-
     - name: Install libudev-dev
       run: sudo apt install -y libudev-dev
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -48,6 +48,12 @@ jobs:
     runs-on: ${{ matrix.os }}
 
     steps:
+    - name: Update packages
+      run: sudo apt update
+
+    - name: Upgrade packages
+      run: sudo apt upgrade -y
+
     - name: Install libudev-dev
       run: sudo apt install -y libudev-dev
 
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -1783,7 +1783,8 @@ if_route(unsigned char cmd, const struct rt *rt)
 	if (rt->rt_lifetime != 0) {
 		uint32_t expires;
 
-		expires = lifetime_left(rt->rt_lifetime, &rt->rt_acquired, NULL);
+		expires = lifetime_left(rt->rt_lifetime, &rt->rt_acquired,
+		    NULL);
 		add_attr_32(&nlm.hdr, sizeof(nlm), RTA_EXPIRES, expires);
 	}
 #endif
--- a/src/if-linux.c
+++ b/src/if-linux.c
@@ -1783,7 +1783,7 @@ if_route(unsigned char cmd, const struct rt *rt)
 	if (rt->rt_lifetime != 0) {
 		uint32_t expires;
 
-		expires = lifetime_left(rt->rt_lifetime, &rt->rt_aquired, NULL);
+		expires = lifetime_left(rt->rt_lifetime, &rt->rt_acquired, NULL);
 		add_attr_32(&nlm.hdr, sizeof(nlm), RTA_EXPIRES, expires);
 	}
 #endif
--- a/src/ipv6.c
+++ b/src/ipv6.c
@@ -2308,7 +2308,7 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 			rt->rt_pref = ipv6nd_rtpref(rinfo->flags);
 #endif
 #ifdef HAVE_ROUTE_LIFETIME
-			rt->rt_aquired = rinfo->acquired;
+			rt->rt_acquired = rinfo->acquired;
 			rt->rt_lifetime = rinfo->lifetime,
 #endif
 			rt_proto_add(routes, rt);
@@ -2325,7 +2325,7 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 				rt->rt_pref = ipv6nd_rtpref(rap->flags);
 #endif
 #ifdef HAVE_ROUTE_LIFETIME
-				rt->rt_aquired = addr->acquired;
+				rt->rt_acquired = addr->acquired;
 				rt->rt_lifetime = addr->prefix_vltime;
 #endif
 
@@ -2361,7 +2361,7 @@ inet6_raroutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx)
 		rt->rt_pref = ipv6nd_rtpref(rap->flags);
 #endif
 #ifdef HAVE_ROUTE_LIFETIME
-		rt->rt_aquired = rap->acquired;
+		rt->rt_acquired = rap->acquired;
 		rt->rt_lifetime = rap->lifetime;
 #endif
 
@@ -2400,7 +2400,7 @@ inet6_dhcproutes(rb_tree_t *routes, struct dhcpcd_ctx *ctx, enum DH6S dstate)
 				continue;
 			rt->rt_dflags |= RTDF_DHCP;
 #ifdef HAVE_ROUTE_LIFETIME
-			rt->rt_aquired = ia->acquired;
+			rt->rt_acquired = ia->acquired;
 			rt->rt_lifetime = ia->prefix_vltime;
 #endif
 			rt_proto_add(routes, rt);
--- a/src/route.c
+++ b/src/route.c
@@ -558,7 +558,7 @@ rt_cmp_lifetime(struct rt *nrt, struct rt *ort)
 	struct timespec ts;
 	uint32_t deviation;
 
-	timespecsub(&nrt->rt_aquired, &ort->rt_aquired, &ts);
+	timespecsub(&nrt->rt_acquired, &ort->rt_acquired, &ts);
 	if (ts.tv_sec < 0)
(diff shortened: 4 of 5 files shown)


squarooticus:
It's very hard to follow the logic of how this code is supposed to work. It looks like:

* When `rt_doroute()` returns true, `rt_build()` replaces the original route (i.e., the one in dhcpcd's internal tables) with the one it acquired from the network (e.g., via DHCP or RA)
* `rt_doroute()` returns true when it indicates either (a) some change in the route (makes sense) and is successful in making that change or (b) no change (???).

I'm struggling to figure out why (b). The reason I'm trying to puzzle this out is that I'm not sure that modifying the newly-acquired route to match the original one is the right solution. I think it'll work, but AFAICT it's the only place where the newly-acquired route is synthetically modified, which feels like the wrong solution here.

It seems like the right class of solution here is to leave the original route in place when doroute thinks nothing has changed, whereupon eventually the acquisition time difference will exceed DEV_MAX, and the route will be replaced, both in the internal tables and the kernel. But to be confident about this, I first need to understand why (b) above. Can any of the core devs for dhcpcd explain that logic?

rsmarples:
> It's very hard to follow the logic of how this code is supposed to work. It looks like:
> 
> * When `rt_doroute()` returns true, `rt_build()` replaces the original route (i.e., the one in dhcpcd's internal tables) with the one it acquired from the network (e.g., via DHCP or RA)
> * `rt_doroute()` returns true when it indicates either (a) some change in the route (makes sense) and is successful in making that change or (b) no change (???).
> 
> I'm struggling to figure out why (b).

We return true on no change so we can update our internal tables.
This is for the case when routes already exist and dhcpcd is just starting up.
We really don't want to touch them if we don't need to. If we do some OS might flush the ARP table and cause connectivity issues with in-use network applications.

> The reason I'm trying to puzzle this out is that I'm not sure that modifying the newly-acquired route to match the original one is the right solution. I think it'll work, but AFAICT it's the only place where the newly-acquired route is synthetically modified, which feels like the wrong solution here.
> 
> It seems like the right class of solution here is to leave the original route in place when doroute thinks nothing has changed, whereupon eventually the acquisition time difference will exceed DEV_MAX, and the route will be replaced, both in the internal tables and the kernel. But to be confident about this, I first need to understand why (b) above. Can any of the core devs for dhcpcd explain that logic?

I think the change makes sense and I've added a supporting comment for it.
@squarooticus does this help?

squarooticus:
It's your code. 🤷‍♂️  I'm mainly worried about adding another layer of complexity to the onion, in a place where there have been subtle logic problems in the past. But I do agree that this PR will likely fix the immediate issue.

squarooticus:
I guess I can outline how I would manage routes if I were writing this from scratch. This is just me opining. Feel free to ignore. ;-)

Essentially, whenever receiving an RA, for the default route:

1. Compare the default route info to our internal state.

    - If there was no change, check the last time I updated this route in the kernel and if it exceeds some configurable interval, atomically replace/add the route (mostly to reset the lifetime) and reset the last-update time to now(). (This is to avoid replacing the route after literally every RA on networks that send them every 1-2 seconds. Maybe this isn't necessary, and it's cheap enough to do every time, but it feels like something worthy of a hack, especially if the expiration time is many multiples of the Adv interval, as it is on most networks.)

    - If there *was* a change (or we just started up and had no previous internal state), atomically replace/add the route with that from the RA, update the internal state to match the RA, and reset the last-update time to now().

2. Clean up any vestigial routes we added that might have been left over, e.g., because of a metric change in the config.

    - This is a little complicated because for proper compartmentalization you don't necessarily want to duplicate all of the kernel's logic for the key fields in a route that it uses to determine replace vs. add. You could instead define two protos and alternate between them for each update, flushing rules from the other proto after doing your atomic replaces. Having a custom proto also makes it easy to clean up all the routes you added on termination.

You can do the same basic thing for RIOs.

PIOs for SLAAC networks are a little more complicated, but the same general idea applies (don't replace too often on networks that flood RAs).

rsmarples:
> I guess I can outline how I would manage routes if I were writing this from scratch. This is just me opining. Feel free to ignore. ;-)
> 
> Essentially, whenever receiving an RA, for the default route:
> 
> 1. Compare the default route info to our internal state.
>    
>    * If there was no change, check the last time I updated this route in the kernel and if it exceeds some configurable interval, atomically replace/add the route (mostly to reset the lifetime) and reset the last-update time to now(). (This is to avoid replacing the route after literally every RA on networks that send them every 1-2 seconds. Maybe this isn't necessary, and it's cheap enough to do every time, but it feels like something worthy of a hack, especially if the expiration time is many multiples of the Adv interval, as it is on most networks.)

This is essentially what we do now.

>    * If there _was_ a change (or we just started up and had no previous internal state), atomically replace/add the route with that from the RA, update the internal state to match the RA, and reset the last-update time to now().

We actively want to avoid this on BSD as any route change will flush the internal ARP tables.

> 2. Clean up any vestigial routes we added that might have been left over, e.g., because of a metric change in the config.
>    
>    * This is a little complicated because for proper compartmentalization you don't necessarily want to duplicate all of the kernel's logic for the key fields in a route that it uses to determine replace vs. add. You could instead define two protos and alternate between them for each update, flushing rules from the other proto after doing your atomic replaces. Having a custom proto also makes it easy to clean up all the routes you added on termination.

Yes a metric change is complicated and is somewhat supported in dhcpcd.
Although I haven't changed it works still for a number of years.

> You can do the same basic thing for RIOs.
> 
> PIOs for SLAAC networks are a little more complicated, but the same general idea applies (don't replace too often on networks that flood RAs).

For dhcpcd it's just a route. The route code is agnostic to its source and works for IPv4 and IPv6.

Thanks for your comments :)

squarooticus:
> > * If there _was_ a change (or we just started up and had no previous internal state), atomically replace/add the route with that from the RA, update the internal state to match the RA, and reset the last-update time to now().
> 
> We actively want to avoid this on BSD as any route change will flush the internal ARP tables.

It flushes the entire ARP table/ND table for all interfaces whenever any route changes? Wow.

> > You can do the same basic thing for RIOs.
> > PIOs for SLAAC networks are a little more complicated, but the same general idea applies (don't replace too often on networks that flood RAs).
> 
> For dhcpcd it's just a route. The route code is agnostic to its source and works for IPv4 and IPv6.

dhcpcd does set `accept_ra=0` on Linux on startup, disabling kernel autoconfiguration, so it must be handling SLAAC itself for PIOs marked Autonomous. (I know it works because my Starlink-facing interface has `accept_ra=0` and that interface still manages to get an address, and also the logs say it's doing that. This is with 10.1.0, at least, so maybe that behavior changed. I decided to downgrade from 10.3.2, but even though I have an exit hook workaround I am not foolish enough to attempt that upgrade from several hundred miles away.)

Anyway, thanks for the quick reply.

rsmarples:
> > > * If there _was_ a change (or we just started up and had no previous internal state), atomically replace/add the route with that from the RA, update the internal state to match the RA, and reset the last-update time to now().
> > 
> > 
> > We actively want to avoid this on BSD as any route change will flush the internal ARP tables.
> 
> It flushes the entire ARP table/ND table for all interfaces whenever any route changes? Wow.

Only for entries attached to the route.
This makes sense when you think about it.

To get the same behaviour (which is important for multi-homed hosts) I need to delete and re-add the route.

> > > You can do the same basic thing for RIOs.
> > > PIOs for SLAAC networks are a little more complicated, but the same general idea applies (don't replace too often on networks that flood RAs).
> > 
> > 
> > For dhcpcd it's just a route. The route code is agnostic to its source and works for IPv4 and IPv6.
> 
> dhcpcd does set `accept_ra=0` on Linux on startup, disabling kernel autoconfiguration, so it must be handling SLAAC itself for PIOs marked Autonomous. (I know it works because my Starlink-facing interface has `accept_ra=0` and that interface still manages to get an address, and also the logs say it's doing that. This is with 10.1.0, at least, so maybe that behavior changed. I decided to downgrade from 10.3.2, but even though I have an exit hook workaround I am not foolish enough to attempt that upgrade from several hundred miles away.)
> 
> Anyway, thanks for the quick reply.

Ah let me rephrase. dhcpcd routing takes routes from many sources - RA PIO is just one, it could also be RA Route Information Option (RIO), or any number of DHCP options such as Classless Static Routing. The routing logic is agnostic to all of this.

Issue #614, opened 2026-05-24
Segfault in `ipv6nd.c`


lima-limon-inc:
I've been having ocasional segfaults on `dhcpcd`. I couldn't pinpoint the issue yet, but I do believe it only happens when I have docker running.
I've enabled a core dump and it appears that the segfault occurred on this line:
```c
	if (ifp->ctx->ra_routers == NULL)
		return;
```
[Link to the file](https://github.com/NetworkConfiguration/dhcpcd/blob/20ec0df64b3dddd6255a7e01c4808e73df9d7be1/src/ipv6nd.c#L451-L452).

Maybe this is related to https://github.com/NetworkConfiguration/dhcpcd/issues/566

OS: Gentoo Linux
Arch: `x86_64`
dhcpcd version: `10.3.2`

09F911029D74E35BD84156C5635688C0:
Not sure what the aislop policy is in this project but I made this based on the code reference above and it seems to have worked.

```
mkdir -p /etc/portage/patches/net-misc/dhcpcd
cp dhcpcd-10.3.2-fix-ipv6nd-expire-uaf.patch /etc/portage/patches/net-misc/dhcpcd/
emerge -av1 net-misc/dhcpcd
```

[dhcpcd-10.3.2-fix-ipv6nd-expire-uaf.patch](https://github.com/user-attachments/files/30793199/dhcpcd-10.3.2-fix-ipv6nd-expire-uaf.patch)

> ## History
> * cf94e2dd (v8.0.0, 2019): `ipv6nd_startexpire()` introduced; timer in the default queue, so `ipv6nd_drop()`'s generic delete covered it (except the `ra_routers == NULL` early-return path).
> * 3e0c93a4 / 3c9cea4b (v9.0.0, 2020): timer moved to the dedicated `ELOOP_IPV6RA_EXPIRE` queue; the queue-scoped delete in `ipv6nd_drop()` silently stopped matching it. Dangling timer possible from here on.
> * 2a3fb4ef (v10.3.1, 2026, "Fixes #584, Helps #566"): added the correct queue deletion, but in `ipv6nd_removefreedrop_ra()`, which never executes for interfaces with zero RAs — leaving exactly the case above unfixed. Consistent with reports persisting on 10.3.1/10.3.2 (also https://bugs.gentoo.org/966588).

Appears still not fixed in 10.5.0 but patch works.

Cheaterman:
[...]

Issue #566, opened 2025-12-11
dhcpcd 10.3.0 segmentation fault in interaction with docker


adavidzh:
Not present without `docker` running.

Not present in `10.2.4`.

Issue opened here for reference and in case others are affected.

More information under https://bugs.gentoo.org/966588

rsmarples:
Can you test the commit on the above branch please to see if that fixes it for you?

moabeat-berlin:
I still had the issue of segfaulting in combination with docker using version 10.3.2 of dhcpcd. On my server a docker container was constantly restarting, approximately every 10 seconds. This resulted in virtual network interfaces removed and re-created frequently, which was visible in syslog. In this case the segfault of dhcpcd happened approximately 25h and 30m later. I could imagine the issue might arise sooner the more virtual network interfaces are removed and re-created. Calculating the time-frame and restart frequency, the issue should occur after roughly 9,000 - 10,000 container restarts.

This actually fits well to the [bug report submitted at Gentoo](https://bugs.gentoo.org/966588). The issue arises when containers are frequently created and removed, like when building docker images or running CI agents.

Cheaterman:
I believe this isn't quite #614 after all ; I just had a segfault today despite using the attached user epatch.

dmesg:
```
[1623412.783682] dhcpcd[1815582]: segfault at 14 ip 00005639c062b40d sp 00007ffdabfc2d80 error 4 in dhcpcd[3240d,5639c0601000+3e000] likely on CPU 11 (core 3, socket 0)
[1623412.783704] Code: 00 4d 85 ed 0f 84 53 02 00 00 41 f6 85 a0 00 00 00 01 0f 84 45 02 00 00 c5 fd 6f 1c 24 c5 fd 7f 5c 24 60 49 8b 85 98 00 00 00 <8b> 40 14 ff c8 83 f8 fd 0f 87 65 05 00 00 48 89 df c5 f8 77 e8 5a
```

[dhcpcd-10.3.2-fix-ipv6nd-expire-uaf.patch](https://github.com/user-attachments/files/31065402/dhcpcd-10.3.2-fix-ipv6nd-expire-uaf.patch)

rsmarples:
@Cheaterman Check the commit 0c1f40749 for the full fix, it does more than your patch.

Cheaterman:
Thanks for the heads up, I applied the attached user epatch instead, it definitely has a better shape 👍 

[...]

PR #698, opened 2026-08-12
ipv6nd: don't leave a carrier expiry timeout on a freed interface

Hojun-Cho:
An attempt at #614.

dhcpcd faults about 17 seconds after a veth gains carrier and is then
removed, but only when that interface never received an RA. On this
machine dhcpcd 10.3.2 hits it whenever a docker container starts, which
is the same version and trigger as #566.

Reproducer:

```sh
#!/bin/sh
# unshare -rn sh repro.sh /path/to/dhcpcd/src/dhcpcd
#
# build it with its own directories or it will fight the dhcpcd already
# running on the machine over the pidfile, control socket and duid:
#   d=$(mktemp -d)
#   CC=clang ./configure --sanitize --disable-privsep \
#       --rundir="$d/run" --dbdir="$d/db"

ip link set lo up
ip link add veth0 type veth peer name veth1	# left down on purpose

"$1" -d -B --script=/bin/true --nohook=all --noipv4 veth0 &
pid=$!
sleep 4

ip link set veth1 up
ip link set veth0 up		# carrier acquired, ipv6nd_startexpire()
sleep 3

ip link del veth0		# interface departed, if_free(ifp)
sleep 20			# RTR_CARRIER_EXPIRE is 17

kill $pid 2>/dev/null
wait $pid
```

The veth has to start down or dhcpcd never sees the carrier transition
and `dhcpcd_handlecarrier()` returns early on `was_link_up`.

On master (7c87dfdf), from a git checkout so `--sanitize` is not
silently dropped:

```
==11428==ERROR: AddressSanitizer: heap-use-after-free
READ of size 8
    #0 ipv6nd_expire src/ipv6nd.c:447
    #1 eloop_start src/eloop.c:1167

freed by thread T0 here:
    #1 dhcpcd_handleinterface src/dhcpcd.c:1096
    #4 if_handlelink src/if-linux.c:1196

previously allocated by thread T0 here:
    #1 if_discover src/if.c:570
```

That is `ifp->ctx->ra_routers` at the top of `ipv6nd_expire()`, same as
the backtrace in gentoo 966588 comment 30.

`ipv6nd_startexpire()` guards on `ra_routers == NULL`, but `ipv6_init()`
allocates that head as an empty TAILQ, so the guard passes even when the
interface has no routers and the timeout is added anyway. It then
outlives `if_free()`. `ipv6nd_free()` needs the cancel as well, since an
interface that is no longer active skips `stop_interface()` when it
departs (`dhcpcd.c:1091`).

The cancel 2a3fb4ef added to `ipv6nd_removefreedrop_ra()` is never
reached in that case, and taking out an interface wide timeout from a
per router path also stops the carrier expiry of the interface's other
routers, so I dropped it. ASan is clean either way when an RA is
accepted before the interface is deleted, so keep the line if you
prefer.

The reproducer is clean after the patch with and without an RA, and the
CI matrix builds and passes `make tests` locally.

### DIFF
--- a/src/ipv6nd.c
+++ b/src/ipv6nd.c
@@ -454,17 +454,31 @@ ipv6nd_expire(void *arg)
 	ipv6nd_expirera(ifp);
 }
 
+static void
+ipv6nd_cancelexpire(struct interface *ifp)
+{
+	eloop_q_timeout_delete(ifp->ctx->eloop, ELOOP_IPV6RA_EXPIRE,
+	    ipv6nd_expire, ifp);
+}
+
 void
 ipv6nd_startexpire(struct interface *ifp)
 {
 	struct ra *rap;
+	bool found = false;
 
 	if (ifp->ctx->ra_routers == NULL)
 		return;
 
 	TAILQ_FOREACH(rap, ifp->ctx->ra_routers, next) {
-		if (rap->iface == ifp)
+		if (rap->iface == ifp) {
 			rap->willexpire = true;
+			found = true;
+		}
+	}
+	if (!found) {
+		ipv6nd_cancelexpire(ifp);
+		return;
 	}
 	eloop_q_timeout_add_sec(ifp->ctx->eloop, ELOOP_IPV6RA_EXPIRE,
 	    RTR_CARRIER_EXPIRE, ipv6nd_expire, ifp);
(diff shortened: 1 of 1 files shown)


src/ipv6nd.c
```diff
@@ -707,6 +719,7 @@ ipv6nd_free(struct interface *ifp)
 	struct dhcpcd_ctx *ctx;
 	ssize_t n;
 
+	ipv6nd_cancelexpire(ifp);
```
rsmarples: This could be moved to line 727, just before the `#ifdef __sun` block yes?

rsmarples (commented):
This is a very nice patch, thanks!

What do you think of my one comment?

src/ipv6nd.c
Hojun-Cho: Thanks. Yes, moved. With the `found` guard no timeout can be outstanding when `state` is NULL anyway, since an RA needs `rs_state`.
Retested, clean with and without an RA.

PR #718, opened 2026-09-01
Add support for RFC 2132 Message option in NAK handling

ColinMcInnes:
If the dhcpcd server sends Option 56 (message) as part of a DHCPv4 NAK, pass it on to the exit hook environment so it can be acted on if necessary.

### DIFF
--- a/src/dhcp.c
+++ b/src/dhcp.c
@@ -3274,7 +3274,7 @@ dhcp_handledhcp(struct interface *ifp, struct bootp *bootp, size_t bootp_len,
 			if (state->message)
 				free(state->message);
 			state->message = get_option_string(ifp->ctx, bootp,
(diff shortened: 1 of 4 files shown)


src/dhcp.c
```diff
@@ -3271,6 +3271,10 @@ dhcp_handledhcp(struct interface *ifp, struct bootp *bootp, size_t bootp_len,
 		if (state->state == DHS_INFORM) /* INFORM should not be NAKed */
 			return;
 		if (!(ifp->ctx->options & DHCPCD_TEST)) {
+			if (state->message)
+				free(state->message);
+			state->message = get_option_string(ifp->ctx, bootp,
```
ColinMcInnes, line 3276: I'm not clear why clang is upset with this

PR #727, opened 2026-09-08
dhcpcd: size the escaped-SSID buffers for the worst case

### LINKED ISSUE #726: dhcpcd_selectprofile: No buffer space available
Associating with a network whose SSID contains non-ASCII characters:

```
wlan0: connected to Access Point: WPA3\354\231\200\354\235\264\355\214\214\354\235\264\347\204\241\347\267\232\344\270\212\347\266\262WiFi
dhcpcd_selectprofile: No buffer space available
```

The SSID is deliberately 32 bytes, with UTF-8 (a test from a different project, but useful here!).

Tracing through the code `dhcpcd_selectprofile()` escapes the SSID into `char pssid[PROFILE_LEN]`, `PROFILE_LEN` is 64, `print_string()` renders each non-printable byte as \NNN (four characters), and an SSID may be `IF_SSIDLEN`  (32 bytes). The worst case therefore needs 32 * 4 + 1 = 129 bytes.

This is not a security issue, nothing is written past the end of either buffer. pssid is NUL-terminated on every path.

`dhcpcd_reportssid()` escapes the same ssid into a buffer sized differently (and explains why the reported ssid is visible in the log line):

```
  dhcpcd.c:720  dhcpcd_reportssid()     char pssid[IF_SSIDLEN * 4];   128 - OK
```

From the maths above, this allocation is one byte short for the trailing NUL; again not a security issue.

akiernan:
print_string() renders each non-printable byte as \NNN - four characters plus terminating NUL, failing with ENOBUFS if there is no room for it. An SSID is up to IF_SSIDLEN (32) bytes, so an escaped SSID needs up to (IF_SSIDLEN * 4) + 1 = 129 bytes.

dhcpcd_selectprofile() uses PROFILE_LEN, which is 64 causing:

  dhcpcd_selectprofile: No buffer space available

in the log, the resulting call to read_config() then gets an an empty SSID, so no `profile ssid ...` block matches.

dhcpcd_reportssid() uses IF_SSIDLEN * 4, which is 128 and correct except for the NUL, so it fails only on a 32-byte SSID whose every byte escapes. It then logs an error instead of the "connected to Access Point" line.

Closes #726


ColinMcInnes (approved):
I agree, no security hole here, just a truncated log, and this should resolve that.

src/dhcpcd.c
```diff
@@ -628,7 +628,7 @@ int
 dhcpcd_selectprofile(struct interface *ifp, const char *profile)
 {
 	struct if_options *ifo;
-	char pssid[PROFILE_LEN];
+	char pssid[(IF_SSIDLEN * 4) + 1];
```
rsmarples, line 631: Why not do this in dhcpcd.h
```
#define IF_SSIDSTRLEN ((IF_SSIDLEN * 4) + 1)
#define PROFILE_LEN IF_SSIDSTRLEN
```
Then you could re-use IF_SSIDSTRLEN in `dhcpcd_reportssid`

rsmarples (commented):
This looks like a good improvement, but what do you think of my suggestion?

src/dhcpcd.c
akiernan: Unless I'm missing something PROFILE_LEN doesn't need to be the same size as an expanded SSID? Though clearly IF_SSIDSTRLEN is a better answer and I found what looked like three more place where it would be used (PR updated).

ColinMcInnes: We could break it out to make a specific PSSID len. But either way it should be defined in dhcpcd.h instead of magic math'd in the function.

rsmarples (changes requested):
This now looks good, just fix the formatting please.
You can run `make format` with clang-format v21 installed to fix.

ColinMcInnes (approved):

rsmarples:
Thanks!

PR #733, opened 2026-09-14
privsep: Fix daemonising broken by RLIMIT_NOFILE of 0

### LINKED ISSUE #716: [Regression] stdout/stderr handling in 10.5.2
Hello,

on 10.3.2, for all processes dhcpcd creates

i see in `/proc/` that stdout/stderr are `/dev/null`:

```sh
ls -lah /proc/<pid>/fd
...
l-wx------    1 root     root          64 Aug 28 16:58 1 -> /dev/null
l-wx------    1 root     root          64 Aug 28 16:58 2 -> /dev/null
```

However with 10.5.2 i see

```sh
l-wx------    1 root     root          64 Aug 27 18:20 1 -> pipe:[54494]
l-wx------    1 root     root          64 Aug 27 18:20 2 -> pipe:[54495]
```

for the initial process (started via a script), or

```sh
lrwx------    1 root     root          64 Aug 27 18:52 1 -> /dev/ttymxc2
lrwx------    1 root     root          64 Aug 27 18:52 2 -> /dev/ttymxc2
```

when called more directly for testing purposes.

That has caused issues in our deployment due to some other process ending up in a blocking read on the pipe. Used to work fine up to 10.3.2 inclusively.
Now i tried to follow the code and i assume for some reason `dhcpcd_daemonise` hasn't been called.

Parameters for dhcpcd are `-b -qq -f </path/to/the/config> <interface>`.

jcronenberg:
## Problem
ps_dropprivs() sets RLIMIT_NOFILE to {0,0} after dropping privileges. On Linux dup2(oldfd, newfd) fails EBADF once newfd >= RLIMIT_NOFILE, even for an already-open fd, so dhcpcd_daemonised()'s dup2 onto stdout/stderr silently stops working (the return value isn't checked). Every daemonised process then keeps holding onto whatever stdio it inherited at fork forever, which hangs anything reading from a piped stdout/stderr waiting for EOF that never comes.

Reproducer: `dhcpcd --ipv4only --waitip --persistent --noarp eth0 | cat` applies the lease but never returns.

Bisected to 62018896, which dropped the NetBSD/DragonFly/kqueue/epoll-only guard around the setrlimit() and made it unconditional, enabling it on Linux for the first time.

## Solution
Fix: cap RLIMIT_NOFILE at STDERR_FILENO + 1 instead of 0. Still blocks new fds - 0-2 are always open, so there is no free slot below the limit to allocate - just leaves 0-2 dup2-able.

Should fix #716 I think


hendrikdonner:
I can confirm that f8959a3666149b5133b2f29b7f216b6c7cc9b808 fixes #716 . Not sure if the `RLIMIT_NOFILE` has other side effects.

src/privsep.c
```diff
@@ -159,10 +159,13 @@ ps_dropprivs(struct dhcpcd_ctx *ctx)
 	struct rlimit rzero = { .rlim_cur = 0, .rlim_max = 0 };
 
 #ifndef __sun /* RLIMIT_NOFILE and ppoll don't mix */
```
rsmarples, line 161: Might be worth adding `__linux__` back to this instead to say that this causes dup2 to fail.
I need RLIMIT_NOFILE of zero for Dragonfly and NetBSD - that is not negotiable.

rsmarples (changes requested):

thresheek:
I can also confirm f8959a3666149b5133b2f29b7f216b6c7cc9b808  fixes the issue as described in https://github.com/NetworkConfiguration/dhcpcd/issues/737

